PR #8, opened 2015-10-08
Refactoring ProfileManager class to try to create the profile folder just before try to access it

viniciusboson:
Refactoring ProfileManager class to try to create the profile folder  if necessary, just before try to access it. It is better than try to create on the constructor because first, if the server does not have permission to do this, the profile manager module will not be loaded, and second even if the folder exists at constructor time, it may be deleted in a future moment and it would fail when access the folder. 

In addition, there are some java codestyle corrections


zestyping:
This is better done in the buendia-server package at installation time (which has been merged in elsewhere), so closing this PR.

viniciusboson:
@zestyping, 

I think this code complements the installation script and it is also util.
We should considere that the folder could not exist, even if it was created on installation time.

zestyping:
Unfortunately it won't work to create the folder in the web server — the
tomcat user account will not have permission to create it, and it should
not be created by that user account anyway.

We don't need to defend against the removal of arbitrary files that are
part of the installation.  We can rely on the package being installed.

PR #10, opened 2015-10-12
Clear the concept_word and concept_reference_* tables

zestyping:
These tables are huge and we don't need them.  Clearing them significantly speeds up installation/flashing of the server software.


dancunningham:
Looks good to me, merging this!

PR #11, opened 2015-10-18
Make orders editable and deletable.

zestyping:


dancunningham:
@zestyping can you make a corresponding pull request on the client app for this?

dancunningham:
Found it! https://github.com/projectbuendia/client/commit/fb2949bb8c6d3764b46269f56cffb97f25c8f104

It's part of the scratch/chad-bokoro branch. @llvasconcellos is going to try and merge it along with some of the other things in that branch.

llvasconcellos:
I merged dev into this and applied changes to make orders editable and "deletable" according to the most recent context.
It's working for edit and delete observations.
@capnfabs Can you have a look at this PR and see if I did not alter anything you worked on the past PRs? There were a lot of your commits to OrderResource.java. I did my best to keep only to change what we needed to make this work.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
```diff
@@ -341,6 +409,30 @@ private Order reviseOrder(Order order, SimpleObject edits) {
                     }
                     break;
 
+                case "start_millis":
```
llvasconcellos: @capnfabs On Ping's initial commit for this functionality he removed the case "stop" from the switch. I kept it. Do we need it?

capnfabs:
Hmmm. This is going to need a little more thought if we want it to work with our current sync model.

capnfabs:
I think there's some good reasons behind why Orders in OpenMRS were modelled like this. Editing orders should only be supported if the order hasn't been executed before, and is fairly recent. Otherwise, it's important to display the old order as well as the new order, for the cases when the user _isn't_ just editing a typo, but is changing a dosage or something.

So, here's my proposal:

- Only allow edits and deletions if less than a day has passed and there's no executions of the order. When an edit occurs, the old order is voided and a new order is created in its stead.
- Beyond that, allow revisions. When a revision occurs, we keep both the old order and the new order. Both of these are displayed on the client.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
capnfabs: It looks like Ping changed the field name from "stop" to "stop_millis" here, so it's ok to remove the "stop".

dancunningham:
Closing as this is superseded by #48

PR #13, opened 2015-10-20
Changing buendia's configuration folder from '/usr/share/' to '/usr/local/opt'  

viniciusboson:
The buendia's configuration file under `/usr/share/` folder should conflict with Security policies like Apple's System Integrity Protection on 'OS X El Capitan' which prevents even users with root privilege from creating folders under `/usr`(except under `/usr/local`).

### DIFF
--- a/devices/server/setup
+++ b/devices/server/setup
@@ -66,8 +66,8 @@ echo $flashed
 rm -rf /tmp/debian
 mkdir /tmp/debian
 if [ -n "$site_id" ]; then
-    mkdir -p /tmp/debian/usr/share/buendia/site
-    echo "SITE_ID=$site_id" > /tmp/debian/usr/share/buendia/site/id
+    mkdir -p /tmp/debian/usr/local/opt/buendia/site
+    echo "SITE_ID=$site_id" > /tmp/debian/usr/local/opt/buendia/site/id
 fi
 if [ -n "$repo_url" ]; then
     mkdir -p /tmp/debian/etc/apt/sources.list.d
--- a/devices/server/yocto/usr/local/bin/install-buendia-base
+++ b/devices/server/yocto/usr/local/bin/install-buendia-base
@@ -115,7 +115,7 @@ apt-get update
 apt-get -y --allow-unauthenticated install buendia-update
 
 # If a site has been selected, install the site settings package.
-. /usr/share/buendia/site/id || true
+. /usr/local/opt/buendia/site/id || true
 if [ -n "$SITE_ID" ]; then
     apt-get -y --allow-unauthenticated install buendia-site-$SITE_ID
     echo "Buendia site ID configured: $SITE_ID"
--- a/openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
+++ b/openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
@@ -49,7 +49,7 @@ import javax.servlet.http.HttpServletResponse;
 @Controller
 public class ProfileManager {
     static Log log = LogFactory.getLog(ProfileManager.class);
-    final File PROFILE_DIR = new File("/usr/share/buendia/profiles");
+    final File PROFILE_DIR = new File("/usr/local/opt/buendia/profiles");
     final String VALIDATE_CMD = "buendia-profile-validate";
     final String APPLY_CMD = "buendia-profile-apply";
 
--- a/openmrs/omod/src/main/resources/config.xml
+++ b/openmrs/omod/src/main/resources/config.xml
@@ -44,7 +44,7 @@
     <description>
       Name of the currently active profile, which defines the forms and UI customizations for the
       Buendia app.
-      This should be the name of a file in /usr/share/buendia/profiles (including the .csv
+      This should be the name of a file in /usr/local/opt/buendia/profiles (including the .csv
       extension).
     </description>
   </globalProperty>
@@ -87,4 +87,3 @@
   </servlet>
 
 </module>
-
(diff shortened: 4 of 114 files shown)


zestyping:
If this is really a problem then it doesn't make sense to move just one part of the buendia folder; we should move the entire /usr/share/buendia folder instead to e.g. /usr/local/buendia.

Can we confirn that this is an actual conflict though? I've never had any problems with this on my Mac.  Is there any technical documentation on this new feature in El Capitan?

zestyping:
If true it would affect a LOT of other software out there besides Buendia, so it's a bit surprising.

llvasconcellos:
Ping I'm working on a more in depth fix to this problem. And yes, I think we should review all the server folder structure and have one for local instalation (home folder) and other as a service (installed as root).

llvasconcellos:
A good discussion about it is here: http://stackoverflow.com/questions/1510104/where-to-store-application-data-non-user-specific-on-linux

viniciusboson:
It's true, @zestyping.
It's called System Integrity Protection.
You can see this on [wikipedia](https://en.wikipedia.org/wiki/System_Integrity_Protection) or [apple's developer site](https://developer.apple.com/library/prerelease/mac/releasenotes/MacOSX/WhatsNewInOSX/Articles/MacOSX10_11.html#//apple_ref/doc/uid/TP40016227-SW1).

I didn't have this problem until upgrading to El Captian OS.

Here is what happen when i try to create a folder with root privilege under `/usr/share` folder.

screen shot 2015-10-20 at 16 10 52

And here are the files that El Capitan put in quarantine when it was installed
screen shot 2015-10-20 at 16 17 02

viniciusboson:
I agree with you both. We should consider to change other parts to `/usr/local` too.
I can change this PR to do that.

llvasconcellos:
Other good info:
http://superuser.com/questions/90479/what-is-the-conventional-install-location-for-applications-in-linux

viniciusboson:
I've refactored all references of `/usr/share/buendia` to ùsr/local/opt/buendia`.
But i have some doubts yet, before changing this PR.

The files `/packages/buendia-openmrs/control/postinst`,  `/packages/buendia-openmrs/control/postrm` and `tools/install_omod` refer to the path `/usr/share/tomcat7`. Should we refactor this path too ? 

In addition, how is the proper way to test these changes, @zestyping ?

zestyping:
Hi Vinicius—good catch, thanks!  I read a bit more about this and I understand the issue better now.  For El Capitan we do need to address this.  However, I'm not sure what the best solution is yet.  Can you tell me your reasoning behind the choice of `/usr/local/opt/buendia`?

This is a tricky question.  Our target platform is Debian Linux, and we distribute the Buendia system as Debian packages managed by APT.  These packages are supposed to install into `/usr`.  (`/usr/local` is specifically for files _not_ managed in APT packages.)  So, on Linux, `/usr/share/buendia` does seem like the correct location.  `/usr/share/tomcat7` follows the same pattern, as do many other installed packages.

It is only because of El Capitan that we have a problem with `/usr/share/buendia`.  So, we have a few options:

(a) Make the root directory `/usr/share/buendia` configurable so it can be different between Linux and Mac systems.

(b) Require Mac developers to put a symlink at `/usr/share/buendia/packages` just for development (and turn off SIP temporarily to to so).

(c) Make the `/usr/share/buendia/packages` directory configurable so it can be different between Linux and Mac systems.

(d) Change the logic of the Profile Manager so that it looks in two different places for the `packages` directory and uses whichever one exists.

(e) Change the logic of the Profile Manager so that it uses a configuration variable to determine the path to the `packages` directory.

There are probably more options too.  I will think about this more and share thoughts here.

viniciusboson:
The choice was made based on Linux FileSystem Hierarchy that [i've seen](http://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/opt.html).
Accordingly with that, `/opt` "should be reserved for all the software and add-on packages that are not part of the default installation". The additional `/usr/local`, besides the security policy constraint applied by Apple now, [here](http://www.tldp.org/LDP/Linux-Filesystem-Hierarchy/html/usr.html) maybe a good reason to use it:
"The original idea behind '/usr/local' was to have a separate ('local') '/usr' directory on every machine besides '/usr', which might be just mounted read-only from somewhere else. It copies the structure of '/usr'. These days, '/usr/local' is widely regarded as a good place in which to keep self-compiled or third-party programs. The /usr/local hierarchy is for use by the system administrator when installing software locally. It needs to be safe from being overwritten when the system software is updated. It may be used for programs and data that are shareable amongst a group of hosts, but not found in /usr. Locally installed software must be placed within /usr/local rather than /usr unless it is being installed to replace or upgrade software in /usr."

I see your point of view to use `share` as is Debian packages managed by APT. What about if we use `/usr/local/share` ?

@llvasconcellos has suggested us to parametrize this path as a parameter of `tools/openmrs_setup` and record it into/retrieve it from the buendia's runtime propertie file. However, we haven't realized yet how to deal with the other scripts.

zestyping:
I agree that `/usr/local` is a good place to install when the installation is not managed by APT.  (And indeed, with El Capitan it seems to be the only choice?)  I don't see any advice in FHS that recommends `/usr/local/opt` though.

Either `/usr/local/buendia` or `/usr/local/share/buendia` would be okay. `/usr/local/buendia` is consistent with `/usr/local/mysql`.  I think it depends where we put the `buendia-*` executable scripts.

If the scripts go in `/usr/local/bin/buendia-*` then `/usr/local/share/buendia` is more natural.

If the scripts go under the `buendia` directory then `/usr/local/buendia` is more natural, with scripts in `/usr/local/buendia/bin`, following just the same pattern as `/usr/local/mysql` and `/usr/local/mysql/bin`.

zestyping:
I just saw a mention that `/usr/local/bin` might be already in the default `PATH` on Mac OS systems.  If someone can confirm that is the case, then `/usr/local/bin/buendia-*` is probably a more convenient place to put the scripts.

viniciusboson:
you are correct, @zestyping. `/usr/local/bin` is in the default `PATH`. At least on El Captian, it is.

dancunningham:
Let's look at this next week and perhaps raise an issue to resolve this, so we can close this PR for now.

viniciusboson:
I agree, @dancunningham. We can close this one.
I believe this is already addressed by PR #15.

Issue #12, opened 2015-10-20
Profile folder creation handling


llvasconcellos:
During the setup of a new Buendia Server the profile folder is not being created and results on a NullPointerException when trying to list files inside of a non existent folder.

The bellow stack trace is result of this error:
```java
org.springframework.web.util.NestedServletException: Request processing failed; nested exception is java.lang.NullPointerException
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:656)
	at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:549)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:735)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:848)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:698)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1506)
	at org.openmrs.module.web.filter.ForcePasswordChangeFilter.doFilter(ForcePasswordChangeFilter.java:61)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1494)
	at org.openmrs.module.web.filter.ModuleFilterChain.doFilter(ModuleFilterChain.java:72)
	at org.openmrs.module.xforms.web.XformsFilter.doFilter(XformsFilter.java:69)
	at org.openmrs.module.web.filter.ModuleFilterChain.doFilter(ModuleFilterChain.java:70)
	at org.openmrs.module.web.filter.ModuleFilter.doFilter(ModuleFilter.java:54)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1494)
	at org.openmrs.web.filter.OpenmrsFilter.doFilterInternal(OpenmrsFilter.java:107)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:76)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1494)
	at org.springframework.orm.hibernate3.support.OpenSessionInViewFilter.doFilterInternal(OpenSessionInViewFilter.java:198)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:76)
[...]

PR #14, opened 2015-10-22
Profile Manager Error Handling

llvasconcellos:
In depth refactoring for better error handling and displaying on the Profile Manager.

Other refactoring was made to improve profiles.jsp readability.
Also was introduced the ability to setup a custom path for the profile storage folder.
The custom folder configuration must be in openmrs-runtime.properties under profile_manager.profile_dir.
This changes is related to issue #12.

### DIFF
--- a/openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
+++ b/openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
@@ -16,10 +16,10 @@ import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.openmrs.api.context.Context;
-import org.openmrs.projectbuendia.webservices.rest.ConfigurationException;
 import org.openmrs.projectbuendia.webservices.rest.GlobalProperties;
 import org.springframework.stereotype.Controller;
 import org.springframework.ui.ModelMap;
+import org.springframework.web.bind.annotation.ModelAttribute;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.multipart.MultipartFile;
@@ -33,14 +33,8 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.Date;
-import java.util.List;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import java.util.*;
+import java.util.Map.Entry;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -48,34 +42,70 @@ import javax.servlet.http.HttpServletResponse;
 /** The controller for the profile management page. */
 @Controller
 public class ProfileManager {
-    static Log log = LogFactory.getLog(ProfileManager.class);
-    final File PROFILE_DIR = new File("/usr/share/buendia/profiles");
-    final String VALIDATE_CMD = "buendia-profile-validate";
-    final String APPLY_CMD = "buendia-profile-apply";
+    protected static Log log = LogFactory.getLog(ProfileManager.class);
+    private final String PROFILE_DIR_PATH = "/usr/share/buendia/profiles";
+    private final String VALIDATE_CMD = "buendia-profile-validate";
+    private final String APPLY_CMD = "buendia-profile-apply";
+    private File profileDir;
+
+    @ModelAttribute
+    public void getInitializedMyObject() {
+        Properties config = Context.getRuntimeProperties();
+        String configProfileDir = config.getProperty("profile_manager.profile_dir");
+        if(configProfileDir != null){
+            profileDir = new File(configProfileDir);
+        } else {
+            profileDir = new File(PROFILE_DIR_PATH);
+        }
+    }
 
     @RequestMapping(value = "/module/projectbuendia/openmrs/profiles", method = RequestMethod.GET)
     public void get(HttpServletRequest request, ModelMap model) {
-        List<FileInfo> files = new ArrayList<>();
-        for (File file : PROFILE_DIR.listFiles()) {
-            files.add(new FileInfo(file));
-        }
-        Collections.sort(files, new Comparator<FileInfo>() {
-            public int compare(FileInfo a, FileInfo b) {
-                return -a.modified.compareTo(b.modified);
-            }
-        });
-
-        model.addAttribute("user", Context.getAuthenticatedUser());
-        model.addAttribute("profiles", files);
-        model.addAttribute("currentProfile",
-            Context.getAdministrationService().getGlobalProperty(
-                GlobalProperties.CURRENT_PROFILE));
+        String currentProfile = Context.getAdministrationService().getGlobalProperty(GlobalProperties.CURRENT_PROFILE);
+        model.addAttribute("currentProfile", currentProfile);
         model.addAttribute("authorized", authorized());
+        model = queryParamsToModelAttr(request, model);
+        if(!profileDir.exists()){
+            model.addAttribute("success", false);
+            model.addAttribute("message", "The Profile Manager directory does not exist. It must reside on " +
+                    "/usr/share/buendia/profiles or on a directory specified in openmrs-runtime.properties " +
+                    "file under profile_manager.profile_dir directive.");
+        } else {
+            model.addAttribute("profiles", listProfiles());
+        }
+    }
+
+    private ModelMap queryParamsToModelAttr(HttpServletRequest request, ModelMap model){
+        Map params = request.getParameterMap();
+        Iterator paramIterator = params.entrySet().iterator();
+        while (paramIterator.hasNext()) {
+            Entry param = (Entry) paramIterator.next();
+            String key = (String) param.getKey();
+            String[] value = (String[]) param.getValue();
+            model.addAttribute(key, value[0]);
+        }
+        return model;
+    }
+
+    private List<FileInfo> listProfiles() {
+        List<FileInfo> files = new ArrayList<>();
+        File[] profiles = profileDir.listFiles();
+        if(profiles != null) {
+            for (File file : profileDir.listFiles()) {
+                files.add(new FileInfo(file));
+            }
+            Collections.sort(files, new Comparator<FileInfo>() {
+                public int compare(FileInfo a, FileInfo b) {
+                    return -a.modified.compareTo(b.modified);
+                }
+            });
+        }
+        return files;
     }
 
     public static boolean authorized() {
         return Context.hasPrivilege("Manage Concepts") &&
-            Context.hasPrivilege("Manage Forms");
+                Context.hasPrivilege("Manage Forms");
     }
 
     @RequestMapping(value = "/module/projectbuendia/openmrs/profiles", method = RequestMethod.POST)
@@ -90,7 +120,7 @@ public class ProfileManager {
             String filename = request.getParameter("profile");
             String op = request.getParameter("op");
             if (filename != null) {
-                File file = new File(PROFILE_DIR, filename);
+                File file = new File(profileDir, filename);
                 if (file.isFile()) {
                     model.addAttribute("filename", filename);
                     if ("Apply".equals(op)) {
@@ -108,7 +138,7 @@ public class ProfileManager {
     }
 
     /** Chooses a filename based on the given name, with a "-vN" suffix appended for uniqueness. */
-    String getNextVersionedFilename(String name) {
+    private String getNextVersionedFilename(String name) {
         // Separate the name into a sanitized base name and an extension.
         name = sanitizeName(name);
         String ext = "";
@@ -122,7 +152,7 @@ public class ProfileManager {
         // If "name.ext" exists, it counts as version 1.
         String prefix = name + "-v";
         int highestVersion = 0;
-        for (File file : PROFILE_DIR.listFiles()) {
+        for (File file : profileDir.listFiles()) {
             int version = 0;
             String n = file.getName();
             if (n.equals(name + ext)) {
@@ -130,7 +160,7 @@ public class ProfileManager {
             } else if (n.startsWith(prefix) && n.endsWith(ext)) {
                 try {
                     version = Integer.parseInt(
-                        n.substring(prefix.length(), n.length() - ext.length()));
+                            n.substring(prefix.length(), n.length() - ext.length()));
                 } catch (NumberFormatException e) { }
             }
             highestVersion = Math.max(version, highestVersion);
@@ -145,68 +175,79 @@ public class ProfileManager {
     }
 
     /** Handles an uploaded profile. */
-    // TODO: Give some error message if the profile is not uploaded correctly
-    //       (When tomcat7 didn't have permission to PROFILES_DIR it was failing
-    //        without showing any error message to the user)
-    void addProfile(MultipartHttpServletRequest request, ModelMap model) {
+    private void addProfile(MultipartHttpServletRequest request, ModelMap model) {
         List<String> lines = new ArrayList<>();
         MultipartFile mpf = request.getFile("file");
         if (mpf != null) {
+            String message = "";
+            String filename = mpf.getOriginalFilename();
             try {
                 File tempFile = File.createTempFile("profile", null);
                 mpf.transferTo(tempFile);
-                if (execute(VALIDATE_CMD, tempFile, lines)) {
-                    String filename = getNextVersionedFilename(mpf.getOriginalFilename());
-                    File newFile = new File(PROFILE_DIR, filename);
-                    model.addAttribute("filename", filename);
+                boolean exResult = execute(VALIDATE_CMD, tempFile, lines);
+                if (exResult) {
+                    filename = getNextVersionedFilename(mpf.getOriginalFilename());
+                    File newFile = new File(profileDir, filename);
                     FileUtils.moveFile(tempFile, newFile);
-                    model.addAttribute("success", "add");
+                    model.addAttribute("success", true);
+                    message = "Success adding profile: ";
                 } else {
-                    model.addAttribute("failure", "add");
-                    model.addAttribute("filename", mpf.getOriginalFilename());
-                    model.addAttribute("output", StringUtils.join(lines, "\n"));
+                    model.addAttribute("success", false);
+                    message = "Error adding profile: ";
                 }
             } catch (Exception e) {
+                model.addAttribute("success", false);
+                message = "Problem saving uploaded profile: ";
+                lines.add(e.getMessage());
                 log.error("Problem saving uploaded profile", e);
+            } finally {
+                model.addAttribute("operation", "add");
+                model.addAttribute("message", message + filename);
+                model.addAttribute("filename", filename);
+                model.addAttribute("output", StringUtils.join(lines, "\n"));
             }
         }
     }
 
     /** Applies a profile to the OpenMRS database. */
-    void applyProfile(File file, ModelMap model) {
+    private void applyProfile(File file, ModelMap model) {
         List<String> lines = new ArrayList<>();
         if (execute(APPLY_CMD, file, lines)) {
             setCurrentProfile(file.getName());
-            model.addAttribute("success", "apply");
+            model.addAttribute("success", true);
+            model.addAttribute("message", "Success applying profile: " + file.getName());
         } else {
-            model.addAttribute("failure", "apply");
+            model.addAttribute("success", false);
+            model.addAttribute("message", "Error applying profile: " + file.getName());
             model.addAttribute("output", StringUtils.join(lines, "\n"));
         }
     }
 
     /** Downloads a profile. */
-    public void downloadProfile(File file, HttpServletResponse response) {
+    private void downloadProfile(File file, HttpServletResponse response) {
         response.setContentType("application/octet-stream");
         response.setHeader(
-            "Content-Disposition", "attachment; filename=\"" + file.getName() + "\"");
+                "Content-Disposition", "attachment; filename=\"" + file.getName() + "\"");
         try {
             response.getOutputStream().write(
-                Files.readAllBytes(Paths.get(file.getAbsolutePath())));
+                    Files.readAllBytes(Paths.get(file.getAbsolutePath())));
         } catch (IOException e) {
             log.error("Error downloading profile: " + file.getName(), e);
         }
     }
 
     /** Deletes a profile. */
-    void deleteProfile(File file, ModelMap model) {
+    private void deleteProfile(File file, ModelMap model) {
         if (file.getName().equals(getCurrentProfile())) {
-            model.addAttribute("failure", "delete");
-            model.addAttribute("output", "Cannot delete the currently active profile.");
+            model.addAttribute("success", false);
+            model.addAttribute("message", "Cannot delete the currently active profile.");
         } else if (file.delete()) {
-            model.addAttribute("success", "delete");
+            model.addAttribute("success", true);
+            model.addAttribute("message", "Success deleting profile: " + file.getName());
         } else {
+            model.addAttribute("success", false);
+            model.addAttribute("message", "Error deleting profile: " + file.getName());
             log.error("Error deleting profile: " + file.getName());
-            model.addAttribute("failure", "delete");
         }
     }
 
@@ -214,13 +255,13 @@ public class ProfileManager {
      * Executes a command with one argument, returning true if the command succeeds.
      * Gathers the output from stdout and stderr into the provided list of lines.
      */
-    boolean execute(String command, File arg, List<String> lines) {
+    private boolean execute(String command, File arg, List<String> lines) {
         ProcessBuilder pb = new ProcessBuilder(command, arg.getAbsolutePath());
         pb.redirectErrorStream(true);  // redirect stderr to stdout
         try {
             Process proc = pb.start();
             BufferedReader reader = new BufferedReader(
-                new InputStreamReader(proc.getInputStream()));
+                    new InputStreamReader(proc.getInputStream()));
             String line;
             while ((line = reader.readLine()) != null) {
                 lines.add(line);
@@ -229,26 +270,27 @@ public class ProfileManager {
             return proc.exitValue() == 0;
         } catch (Exception e) {
             log.error("Exception while executing: " + command + " " + arg, e);
+            lines.add(e.getMessage());
             return false;
         }
     }
 
     /** Sanitizes a string to produce a safe filename. */
-    String sanitizeName(String filename) {
+    private String sanitizeName(String filename) {
         String[] parts = filename.split("/");
         return parts[parts.length - 1].replaceAll("[^A-Za-z0-9._-]", " ").replaceAll(" +", " ");
     }
 
     /** Gets the global property for the name of the current profile. */
-    String getCurrentProfile() {
+    private String getCurrentProfile() {
         return Context.getAdministrationService().getGlobalProperty(
-            GlobalProperties.CURRENT_PROFILE);
+                GlobalProperties.CURRENT_PROFILE);
     }
 
     /** Sets the global property for the name of the current profile. */
-    void setCurrentProfile(String name) {
+    private void setCurrentProfile(String name) {
         Context.getAdministrationService().setGlobalProperty(
-            GlobalProperties.CURRENT_PROFILE, name);
+                GlobalProperties.CURRENT_PROFILE, name);
     }
 
     public class FileInfo {
@@ -268,10 +310,14 @@ public class ProfileManager {
         public String getFormattedName() {
             return (name + "                              ").substring(0, 30);
         }
-        public Long getSize() { return size; }
+        public Long getSize() {
+            return size;
+        }
         public String getFormattedSize() {
             return String.format("%7d", size);
         }
-        public Date getModified() { return modified; }
+        public Date getModified() {
+            return modified;
+        }
     }
 }
--- a/openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
+++ b/openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
@@ -33,8 +33,15 @@ import java.io.IOException;
 import java.io.InputStreamReader;
 import java.nio.file.Files;
 import java.nio.file.Paths;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Date;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Properties;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
@@ -58,7 +65,7 @@ public class ProfileManager {
 
     @RequestMapping(value = "/module/projectbuendia/openmrs/profiles", method = RequestMethod.GET)
     public void get(HttpServletRequest request, ModelMap model) {
-        String currentProfile = Context.getAdministrationService().getGlobalProperty(GlobalProperties.CURRENT_PROFILE);
+        String currentProfile = getCurrentProfile();
         model.addAttribute("currentProfile", currentProfile);
         model.addAttribute("authorized", authorized());
         model = queryParamsToModelAttr(request, model);
@@ -103,7 +110,7 @@ public class ProfileManager {
 
     public static boolean authorized() {
         return Context.hasPrivilege("Manage Concepts") &&
-                Context.hasPrivilege("Manage Forms");
+            Context.hasPrivilege("Manage Forms");
     }
 
     @RequestMapping(value = "/module/projectbuendia/openmrs/profiles", method = RequestMethod.POST)
@@ -158,7 +165,7 @@ public class ProfileManager {
             } else if (n.startsWith(prefix) && n.endsWith(ext)) {
                 try {
                     version = Integer.parseInt(
-                            n.substring(prefix.length(), n.length() - ext.length()));
+                        n.substring(prefix.length(), n.length() - ext.length()));
                 } catch (NumberFormatException e) { }
             }
             highestVersion = Math.max(version, highestVersion);
@@ -225,10 +232,10 @@ public class ProfileManager {
     private void downloadProfile(File file, HttpServletResponse response) {
         response.setContentType("application/octet-stream");
         response.setHeader(
-                "Content-Disposition", "attachment; filename=\"" + file.getName() + "\"");
+            "Content-Disposition", "attachment; filename=\"" + file.getName() + "\"");
         try {
             response.getOutputStream().write(
-                    Files.readAllBytes(Paths.get(file.getAbsolutePath())));
+                Files.readAllBytes(Paths.get(file.getAbsolutePath())));
         } catch (IOException e) {
             log.error("Error downloading profile: " + file.getName(), e);
         }
@@ -259,7 +266,7 @@ public class ProfileManager {
         try {
             Process proc = pb.start();
             BufferedReader reader = new BufferedReader(
-                    new InputStreamReader(proc.getInputStream()));
+                new InputStreamReader(proc.getInputStream()));
             String line;
             while ((line = reader.readLine()) != null) {
                 lines.add(line);
@@ -282,13 +289,13 @@ public class ProfileManager {
     /** Gets the global property for the name of the current profile. */
     private String getCurrentProfile() {
         return Context.getAdministrationService().getGlobalProperty(
-                GlobalProperties.CURRENT_PROFILE);
+            GlobalProperties.CURRENT_PROFILE);
     }
 
     /** Sets the global property for the name of the current profile. */
     private void setCurrentProfile(String name) {
         Context.getAdministrationService().setGlobalProperty(
-                GlobalProperties.CURRENT_PROFILE, name);
+            GlobalProperties.CURRENT_PROFILE, name);
     }
 
     public class FileInfo {
--- a/openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
+++ b/openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
@@ -48,15 +48,12 @@ public class ProfileManager {
     private final String APPLY_CMD = "buendia-profile-apply";
     private File profileDir;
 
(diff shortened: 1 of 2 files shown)


llvasconcellos:
This will be difficult to understand diff. I think, to better understand the work is to read the files separately like a before and after compare.

llvasconcellos:
To test the error messages simply do a new dev env setup and see Profile Manager output the problems that it finds regarding things that are missing like the folder, permissions, missing scripts, etc.

llvasconcellos:
To set a custom profile directory now we have the option to set the profile_manager.profile_dir directive in openmrs-runtime.properties: 
```bash
profile_manager.profile_dir=/usr/share/buendia/profiles
```
This is not mandatory. Profile Manager will use the default directory if no custom is set.

zestyping:
Good call on adding the configuration property, Leonardo!  I like this solution to the Mac vs. Linux problem.

I've made a few other comments, please have a look.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,70 @@
...
-    final String VALIDATE_CMD = "buendia-profile-validate";
-    final String APPLY_CMD = "buendia-profile-apply";
+    protected static Log log = LogFactory.getLog(ProfileManager.class);
+    private final String PROFILE_DIR_PATH = "/usr/share/buendia/profiles";
+    private final String VALIDATE_CMD = "buendia-profile-validate";
+    private final String APPLY_CMD = "buendia-profile-apply";
+    private File profileDir;
+
+    @ModelAttribute
+    public void getInitializedMyObject() {
```
zestyping: It's not obvious how or when this method gets called.  How about making this a getter instead, i.e. `public File getProfileDir()`?

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,70 @@
...
+        Properties config = Context.getRuntimeProperties();
+        String configProfileDir = config.getProperty("profile_manager.profile_dir");
```
zestyping: `getProperty()` has a nice feature that it takes a default value as a second argument, so this can be simplified to:

`config.getProperty("profile_manager.profile_dir", PROFILE_DIR_PATH)`

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,70 @@
...
     @RequestMapping(value = "/module/projectbuendia/openmrs/profiles", method = RequestMethod.GET)
     public void get(HttpServletRequest request, ModelMap model) {
-        List<FileInfo> files = new ArrayList<>();
-        for (File file : PROFILE_DIR.listFiles()) {
-            files.add(new FileInfo(file));
+        String currentProfile = Context.getAdministrationService().getGlobalProperty(GlobalProperties.CURRENT_PROFILE);
+        model.addAttribute("currentProfile", currentProfile);
+        model.addAttribute("authorized", authorized());
+        model = queryParamsToModelAttr(request, model);
+        if(!profileDir.exists()){
```
zestyping: Space between `if` and `(`
Space between `)` and `{`

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,70 @@
...
-        });
+    }
 
-        model.addAttribute("user", Context.getAuthenticatedUser());
-        model.addAttribute("profiles", files);
-        model.addAttribute("currentProfile",
-            Context.getAdministrationService().getGlobalProperty(
-                GlobalProperties.CURRENT_PROFILE));
-        model.addAttribute("authorized", authorized());
+    private ModelMap queryParamsToModelAttr(HttpServletRequest request, ModelMap model){
```
zestyping: Space between `)` and `{`

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,70 @@
...
+        Map params = request.getParameterMap();
```
zestyping: A for loop does the job nicely here, and using the parameterized type `Map<String, String[]>` gives better type safety:
```
        for (Map.Entry<String, String[]> entry : params.entrySet()) {
            model.addAttribute(entry.getKey(), entry.getValue()[0]);
        }
```

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,70 @@
...
+            String[] value = (String[]) param.getValue();
+            model.addAttribute(key, value[0]);
+        }
+        return model;
+    }
+
+    private List<FileInfo> listProfiles() {
+        List<FileInfo> files = new ArrayList<>();
+        File[] profiles = profileDir.listFiles();
+        if(profiles != null) {
```
zestyping: Space between `if` and `(`

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,70 @@
...
+            for (File file : profileDir.listFiles()) {
```
zestyping: Line 92 already called `listFiles()` so this can be just `File file : profiles`

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -130,7 +160,7 @@ String getNextVersionedFilename(String name) {
             } else if (n.startsWith(prefix) && n.endsWith(ext)) {
                 try {
                     version = Integer.parseInt(
-                        n.substring(prefix.length(), n.length() - ext.length()));
+                            n.substring(prefix.length(), n.length() - ext.length()));
```
zestyping: 4-space indentation for continuation lines

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +175,79 @@ String getNextVersionedFilename(String name) {
...
+            String message = "";
+            String filename = mpf.getOriginalFilename();
             try {
                 File tempFile = File.createTempFile("profile", null);
                 mpf.transferTo(tempFile);
-                if (execute(VALIDATE_CMD, tempFile, lines)) {
-                    String filename = getNextVersionedFilename(mpf.getOriginalFilename());
-                    File newFile = new File(PROFILE_DIR, filename);
-                    model.addAttribute("filename", filename);
+                boolean exResult = execute(VALIDATE_CMD, tempFile, lines);
```
zestyping: exResult is not used anywhere else, so it is better to write
```
    if (execute(...)) {
        ...
    }
```

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +175,79 @@ String getNextVersionedFilename(String name) {
...
+                if (exResult) {
+                    filename = getNextVersionedFilename(mpf.getOriginalFilename());
```
zestyping: `getNextVersionedFilename(filename)` since line 183 already did `mpf.getOriginalFilename()`

openmrs/omod/src/main/webapp/profiles.jsp
```diff
@@ -2,104 +2,155 @@
...
+    #upload > span  {
+        position: relative;
+    }
+    #upload > span > input[type="file"] {
+        position: absolute;
+        opacity: 0;
+    }
+    #upload > span > input[type="submit"] {
+        position: absolute;
+        margin-left: 30px;
```
zestyping: If you really want to horizontally align the profile name, the button, and the selection list, don't do it by inserting specific numbers of pixels like this; either use a table, or wrap the labels ("Current profile", "Add a profile", "Select a profile") in their own elements with all the same fixed width, e.g. `display: inline-block; min-width: 10em;`

openmrs/omod/src/main/webapp/profiles.jsp
```diff
@@ -2,104 +2,155 @@
...
 <h3>Project Buendia profiles</h3>
 
 <c:if test="${not authorized}">
-  <c:set var="disabledIfUnauthorized" value="disabled"/>
-  <div class="section error message">
-    You don't have the necessary privileges to manage profiles.
-    Please <a href="/openmrs/login.htm">log in</a> as a user authorized to
-    <b>Manage Concepts</b> and <b>Manage Forms</b>.
-  </div>
+    <c:set var="disabledIfUnauthorized" value="disabled"/>
```
zestyping: All the other .jsp (and .html and .xml) files use two spaces for indentation; better to keep the same in this file as well.

openmrs/omod/src/main/webapp/profiles.jsp
```diff
@@ -2,104 +2,155 @@
...
-
-<c:if test="${not empty param.success}">
-  <div class="section message">
-    <c:if test="${param.success == 'add'}">Added profile</c:if>
-    <c:if test="${param.success == 'apply'}">Applied profile</c:if>
-    <c:if test="${param.success == 'delete'}">Deleted profile</c:if>
-    <span class="filename">${fn:escapeXml(param.filename)}</span>.
-  </div>
+<c:if test="${not empty message}">
+    <div class="section <c:if test="${success == 'false'}">error</c:if> message">
```
zestyping: `success` is a boolean, not a String.  So this is better written
```
<c:if test="${not success}">error</c:if>
```
or more simply
```
${success ? '' : 'error'}
```

openmrs/omod/src/main/webapp/profiles.jsp
```diff
@@ -2,104 +2,155 @@
...
+                <input type="submit" name="op" value="Download">
+                <input type="submit" name="op" value="Delete">
+            </form>
+        </c:otherwise>
+    </c:choose>
 </div>
 
+<script type="text/javascript">
+    (function() {
+        var profileQtd = ${fn:length(profiles)};
```
zestyping: It's not clear what `Qtd` stands for... pick a different name?

openmrs/omod/src/main/webapp/profiles.jsp
```diff
@@ -2,104 +2,155 @@
...
+        if(profileQtd > 3){
```
zestyping: Space between `if` and `(`
Space between `)` and `{`

openmrs/omod/src/main/webapp/profiles.jsp
```diff
@@ -2,104 +2,155 @@
...
+        var authorized = ${authorized};
+        if (!authorized) {
+            for (i = 0; i < document.forms.length; i++) {
+                for (j = 0; j < document.forms[i].length; j++) {
+                    document.forms[i][j].disabled = true;
+                }
+            }
+        }
+
+        document.forms["upload"]["file"].onchange = function(event){
```
zestyping: Use single-quotes instead of double-quotes for consistency with other JavaScript.

Space between `)` and `{`

openmrs/omod/src/main/webapp/profiles.jsp
```diff
@@ -2,104 +2,155 @@
...
+            document.forms['upload'].submit();
+        };
+
+        document.forms["profile-form"].onsubmit = function(event){
```
zestyping: This pops up the confirmation prompt for all the buttons (Add, Download, Delete).  Instead of attaching the onsubmit handler in a separate script, how about putting an `onsubmit` attribute on just the `<input>` element for the Delete button?

Space between `)` and `{`

openmrs/omod/src/main/webapp/profiles.jsp
```diff
@@ -2,104 +2,155 @@
...
+                &nbsp;&nbsp;&nbsp;
+                <input type="submit" name="op" value="Apply">
+                <input type="submit" name="op" value="Download">
+                <input type="submit" name="op" value="Delete">
+            </form>
+        </c:otherwise>
+    </c:choose>
 </div>
 
+<script type="text/javascript">
```
zestyping: Just `<script>`, no need for the `type` attribute.

openmrs/omod/src/main/webapp/profiles.jsp
```diff
@@ -2,104 +2,155 @@
...
+    (function() {
+        var profileQtd = ${fn:length(profiles)};
+        if(profileQtd > 3){
+            document.forms["profile-form"]["profile"].size = profileQtd;
+        }
+
+        var authorized = ${authorized};
+        if (!authorized) {
```
zestyping: This works, but it mixes Java, JSTL EL, and JavaScript, relying on the fact that the Java boolean is converted to a string by JSTL and the Java boolean has a toString() that produces "true" or "false", which coincidentally happens to be spelled the same as a JavaScript boolean... is that what you intended?  It causes JavaScript to be emitted into the page even when it isn't used, so I'm wondering if you mean to use ${authorized} in the JSTL way, e.g. 
```
    <c:if test="${not authorized}">
      <script>
        for (var i = 0; i < document.forms.length; i++) {
          ...
        }
      </script>
    </c:if>
```

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Yes. Sorry, I forgot to change the name of the method!

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Great! Done!

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Done.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Old habits...

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: I definitely tried that but this does not work. entrySet returns a Set. 
So I thought that the while was more straightforward.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: oops!

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: How come I did not see this! Done.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Actualy it is and IntelliJ did this. I'm leaving this way.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Actually I did this on purpose. I think it makes code more clear and readable.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Oops, I did it again! Get out of here Britney! :)
Well spotted. Done.

openmrs/omod/src/main/webapp/profiles.jsp
llvasconcellos: Done.

openmrs/omod/src/main/webapp/profiles.jsp
llvasconcellos: Great! This was before I changed params to models. I forgot to change this. Well done.

openmrs/omod/src/main/webapp/profiles.jsp
llvasconcellos: OK. Old habits die hard. Killed this one!

openmrs/omod/src/main/webapp/profiles.jsp
llvasconcellos: Sorry. It's portuguese for quantity (quantidade). Fixed.

openmrs/omod/src/main/webapp/profiles.jsp
llvasconcellos: I did it on purpose. The result is the same but in a more concise way.

openmrs/omod/src/main/webapp/profiles.jsp
llvasconcellos: Actually It's done this way to make html declaration more clean and separate javascript from htrml. Separate looks from behavior. It's a well used technique.
Fixed spaces and quotes.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -33,14 +33,8 @@
...
 import java.nio.file.Paths;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.Date;
-import java.util.List;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import java.util.*;
```
viniciusboson: I would suggest you to import the specific classes.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,68 @@
...
 
     @RequestMapping(value = "/module/projectbuendia/openmrs/profiles", method = RequestMethod.GET)
     public void get(HttpServletRequest request, ModelMap model) {
-        List<FileInfo> files = new ArrayList<>();
-        for (File file : PROFILE_DIR.listFiles()) {
-            files.add(new FileInfo(file));
+        String currentProfile = Context.getAdministrationService().getGlobalProperty(GlobalProperties.CURRENT_PROFILE);
+        model.addAttribute("currentProfile", currentProfile);
+        model.addAttribute("authorized", authorized());
+        model = queryParamsToModelAttr(request, model);
```
viniciusboson: What does this snippet does ? It would be more clear if you comment it.
```
        String currentProfile = Context.getAdministrationService().getGlobalProperty(GlobalProperties.CURRENT_PROFILE);
        model.addAttribute("currentProfile", currentProfile);
        model.addAttribute("authorized", authorized());
        model = queryParamsToModelAttr(request, model);
```

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: I didn't notice that.  Again I blame IntelliJ.  But I  wonder why it did this.  I'll look into that.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,68 @@
...
+    }
 
-        model.addAttribute("user", Context.getAuthenticatedUser());
-        model.addAttribute("profiles", files);
-        model.addAttribute("currentProfile",
-            Context.getAdministrationService().getGlobalProperty(
-                GlobalProperties.CURRENT_PROFILE));
-        model.addAttribute("authorized", authorized());
+    /** This method sends data to jsp in a more consistent way. */
+    private ModelMap queryParamsToModelAttr(HttpServletRequest request, ModelMap model) {
```
viniciusboson: private methods should be after the publics.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,68 @@
...
+        while (paramIterator.hasNext()) {
+            Entry param = (Entry) paramIterator.next();
+            String key = (String) param.getKey();
+            String[] value = (String[]) param.getValue();
+            model.addAttribute(key, value[0]);
+        }
+        return model;
+    }
+
+    private List<FileInfo> listProfiles() {
```
viniciusboson: private methods should be after the publics.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: It's very clear to me what it does and it wasn't me who did it.  I just rearrange the sequence. I'll leave like this.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,68 @@
...
+                }
+            });
+        }
+        return files;
     }
 
     public static boolean authorized() {
         return Context.hasPrivilege("Manage Concepts") &&
-            Context.hasPrivilege("Manage Forms");
+                Context.hasPrivilege("Manage Forms");
```
viniciusboson: Here another auto indentation issue

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -108,7 +136,7 @@ public View post(HttpServletRequest request, HttpServletResponse response, Model
     }
 
     /** Chooses a filename based on the given name, with a "-vN" suffix appended for uniqueness. */
-    String getNextVersionedFilename(String name) {
+    private String getNextVersionedFilename(String name) {
```
viniciusboson: private methods should be after public ones.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
viniciusboson: I agree with @llvasconcellos. It makes the code more clear.
In my opinion, we should keep if clauses side-effect free.
The only function of an if statement is to test whether a condition is true.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +173,79 @@ String getNextVersionedFilename(String name) {
...
 
     /** Applies a profile to the OpenMRS database. */
-    void applyProfile(File file, ModelMap model) {
+    private void applyProfile(File file, ModelMap model) {
         List<String> lines = new ArrayList<>();
         if (execute(APPLY_CMD, file, lines)) {
             setCurrentProfile(file.getName());
-            model.addAttribute("success", "apply");
+            model.addAttribute("success", true);
+            model.addAttribute("message", "Success applying profile: " + file.getName());
```
viniciusboson: It would be better to i12n the message or to put a TODO to do it later.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +173,79 @@ String getNextVersionedFilename(String name) {
...
         } else {
-            model.addAttribute("failure", "apply");
+            model.addAttribute("success", false);
+            model.addAttribute("message", "Error applying profile: " + file.getName());
```
viniciusboson: It would be better to i12n the message or to put a `TODO` to do it later.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +173,79 @@ String getNextVersionedFilename(String name) {
...
             model.addAttribute("output", StringUtils.join(lines, "\n"));
         }
     }
 
     /** Downloads a profile. */
```
viniciusboson: Though you didn't change this line, i would delete it. It's useless.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +173,79 @@ String getNextVersionedFilename(String name) {
...
-    public void downloadProfile(File file, HttpServletResponse response) {
+    private void downloadProfile(File file, HttpServletResponse response) {
         response.setContentType("application/octet-stream");
         response.setHeader(
-            "Content-Disposition", "attachment; filename=\"" + file.getName() + "\"");
+                "Content-Disposition", "attachment; filename=\"" + file.getName() + "\"");
```
viniciusboson: auto Indentation issue

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +173,79 @@ String getNextVersionedFilename(String name) {
...
         try {
             response.getOutputStream().write(
-                Files.readAllBytes(Paths.get(file.getAbsolutePath())));
+                    Files.readAllBytes(Paths.get(file.getAbsolutePath())));
```
viniciusboson: auto indentation issue

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +173,79 @@ String getNextVersionedFilename(String name) {
...
     }
 
     /** Deletes a profile. */
-    void deleteProfile(File file, ModelMap model) {
+    private void deleteProfile(File file, ModelMap model) {
         if (file.getName().equals(getCurrentProfile())) {
-            model.addAttribute("failure", "delete");
-            model.addAttribute("output", "Cannot delete the currently active profile.");
+            model.addAttribute("success", false);
+            model.addAttribute("message", "Cannot delete the currently active profile.");
```
viniciusboson: It would be better to i12n the message or to put a TODO to do it later.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +173,79 @@ String getNextVersionedFilename(String name) {
...
         try {
             response.getOutputStream().write(
-                Files.readAllBytes(Paths.get(file.getAbsolutePath())));
+                    Files.readAllBytes(Paths.get(file.getAbsolutePath())));
         } catch (IOException e) {
             log.error("Error downloading profile: " + file.getName(), e);
         }
     }
 
     /** Deletes a profile. */
```
viniciusboson: useless comment, though it was already here.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +173,79 @@ String getNextVersionedFilename(String name) {
...
+    private void deleteProfile(File file, ModelMap model) {
         if (file.getName().equals(getCurrentProfile())) {
-            model.addAttribute("failure", "delete");
-            model.addAttribute("output", "Cannot delete the currently active profile.");
+            model.addAttribute("success", false);
+            model.addAttribute("message", "Cannot delete the currently active profile.");
         } else if (file.delete()) {
-            model.addAttribute("success", "delete");
+            model.addAttribute("success", true);
+            model.addAttribute("message", "Success deleting profile: " + file.getName());
```
viniciusboson: It would be better to i12n the message or to put a TODO to do it later.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -145,68 +173,79 @@ String getNextVersionedFilename(String name) {
...
         } else {
+            model.addAttribute("success", false);
+            model.addAttribute("message", "Error deleting profile: " + file.getName());
```
viniciusboson: It would be better to i12n the message or to put a TODO to do it later.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -214,13 +253,13 @@ void deleteProfile(File file, ModelMap model) {
...
      */
-    boolean execute(String command, File arg, List<String> lines) {
+    private boolean execute(String command, File arg, List<String> lines) {
         ProcessBuilder pb = new ProcessBuilder(command, arg.getAbsolutePath());
         pb.redirectErrorStream(true);  // redirect stderr to stdout
         try {
             Process proc = pb.start();
             BufferedReader reader = new BufferedReader(
-                new InputStreamReader(proc.getInputStream()));
+                    new InputStreamReader(proc.getInputStream()));
```
viniciusboson: auto indentation issue

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -229,26 +268,27 @@ boolean execute(String command, File arg, List<String> lines) {
...
         String[] parts = filename.split("/");
         return parts[parts.length - 1].replaceAll("[^A-Za-z0-9._-]", " ").replaceAll(" +", " ");
     }
 
     /** Gets the global property for the name of the current profile. */
-    String getCurrentProfile() {
+    private String getCurrentProfile() {
         return Context.getAdministrationService().getGlobalProperty(
-            GlobalProperties.CURRENT_PROFILE);
+                GlobalProperties.CURRENT_PROFILE);
```
viniciusboson: auto indentation issue

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -229,26 +268,27 @@ boolean execute(String command, File arg, List<String> lines) {
...
     }
 
     /** Sets the global property for the name of the current profile. */
-    void setCurrentProfile(String name) {
+    private void setCurrentProfile(String name) {
         Context.getAdministrationService().setGlobalProperty(
-            GlobalProperties.CURRENT_PROFILE, name);
+                GlobalProperties.CURRENT_PROFILE, name);
```
viniciusboson: auto indentation issue

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
zestyping: I think IntelliJ will get it right if you use the project settings file here: https://github.com/projectbuendia/buendia/blob/dev/intellij-idea-settings.jar

Instructions on how to import it are here: https://github.com/projectbuendia/buendia/wiki/Java-Style

Could you give it a try and see if it fixes the indentation to match the style guide?

viniciusboson:
@llvasconcellos, did you import the project's settings (android-studio-settings.jar)?
It may avoid those auto indentation issues.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
zestyping: Okay.  I guess I'll provide my reasoning for why I think it's not a good way to spend a line of code.

1. It makes the code harder to read because you have to say in your mind "We are going to `execute(...)` and store the success flag in `exResult`, and then we are going to test `exResult`, and if it is true then..." instead of "If `execute(...)` succeeds then..."  That's a lot more thinking.

2. It makes the code harder to change because if you edit the `execute(...)` call, you don't know whether, by changing the meaning of `exResult`, you have affected the behaviour of the program many lines down below at a later point where it happens to use `exResult` again.  (Or, in other words, the variable `exResult` has excessively large scope.)  If you write the expression without `exResult`, it is obvious that editing the `execute(...)` call affects this one `if` statement and cannot affect anything else.

3. Being able to tell whether a line of code has a side effect is important, I agree.  But that is not the responsibility of the `if` statement; it is the responsibility of the method name, `execute`, which is pretty clear that it has a side effect.

I don't think it's a really big deal in this case, but I did want to at least share my thinking with you so you understand where I'm coming from and have a chance to contemplate the benefits of the other way as well.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
zestyping: There's a setting in IntelliJ for this under Preferences > Code Style > Java on the "Imports" tab, named "Class count to use import with '*'".

The settings file at https://github.com/projectbuendia/buendia/blob/dev/intellij-idea-settings.jar should fix this too — it sets the class count to 99 so that `import` with `*` is basically never used.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
```diff
@@ -48,34 +42,68 @@
...
+        String configProfileDir = config.getProperty("profile_manager.profile_dir", PROFILE_DIR_PATH);
+        profileDir = new File(configProfileDir);
+    }
 
     @RequestMapping(value = "/module/projectbuendia/openmrs/profiles", method = RequestMethod.GET)
     public void get(HttpServletRequest request, ModelMap model) {
-        List<FileInfo> files = new ArrayList<>();
-        for (File file : PROFILE_DIR.listFiles()) {
-            files.add(new FileInfo(file));
+        String currentProfile = Context.getAdministrationService().getGlobalProperty(GlobalProperties.CURRENT_PROFILE);
```
zestyping: I just noticed, although you didn't change this either, you could use `getCurrentProfile()` here if you want.

zestyping:
Thanks for the quick response, Leonardo!

viniciusboson:
@llvasconcellos, combining these two PRs  #14 and #15, the profile was successfully imported on a mac environment set up from scratch.

llvasconcellos:
@viniciusboson No, I did not import the settings. I'm doing it right now. Thanks for reminding me and thanks for your input on the Mac side of the force!

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: @zestyping Yes. It fixed indentation. I'll fix rest indentations for the rest of the class.
One thing though, can you add "Show Line Numbers" to your settings file?

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: OK. I fixed it.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: I disagree on this one. Think methods are best placed next to where it is used.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Fixed.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: The entire OpenMRS side of things aren't localized so I don't think it applies here.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: I'm leaving to keep comments consistency.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Fixed.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Fixed

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Fixed.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Fixed.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Fixed.

openmrs/omod/src/main/java/org/projectbuendia/openmrs/web/controller/ProfileManager.java
llvasconcellos: Done!

PR #15, opened 2015-10-22
Fix dev setup script

llvasconcellos:
Added creation of files and folders and setting permissions to proper functioning of Profile Manager on developer machines.

### DIFF
--- a/tools/openmrs_setup
+++ b/tools/openmrs_setup
@@ -71,6 +71,35 @@ mysql -uroot -p <<< "
 "
 echo 'Added MySQL user "openmrs_user" with password "openmrs".'
 
+# Running from Linux or Mac? Set the appropriate path.
+OS=`uname`
+if [[ "$OS" == 'Linux' ]]; then
+    buendia_dir='/usr/share/buendia'
+    bin_dir='/usr/bin'
+elif [[ "$OS" == 'Darwin' ]]; then
+    buendia_dir='/usr/local/opt/buendia'
+    bin_dir='/usr/local/bin'
+fi
+profiles_dir="$buendia_dir/profiles"
+
+
+# Create directories and copy scripts
+sudo mkdir -p $profiles_dir
+sudo cp -p tools/profile_validate "$bin_dir/buendia-profile-validate"
+sudo cp -p packages/buendia-utils/data/usr/share/buendia/utils.sh "$buendia_dir/utils.sh"
+
+# in the profile_apply script there is a hardcoded reference to utils.sh. sed will replace this a mac friendly url.
+# TODO: review this after system wide path related changes.
+if [[ "$OS" == 'Darwin' ]]; then
+    sed 's/share/local\/opt/g' tools/profile_apply > "$bin_dir/buendia-profile-apply"
+else
+    sudo cp -p tools/profile_apply "$bin_dir/buendia-profile-apply"
+fi
+
+# Set permissions
+sudo chmod -R 777 $buendia_dir
+
+
 # Configure OpenMRS to use the openmrs_user MySQL account.
 cat <<EOF >$server_dir/openmrs-runtime.properties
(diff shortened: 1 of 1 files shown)


zestyping:
Yes, it's a great idea to make sure the `profiles` directory exists.  Can we do this without `sudo` though?  At the moment any developer can get going without having root access, and now that you've made the profiles directory configurable it seems like the perfect opportunity to switch away from using a directory that only root can access.  Perhaps something under `$server_dir` since that's already being set up here?

tools/openmrs_setup
```diff
@@ -71,6 +71,33 @@ mysql -uroot -p <<< "
...
+
+# Create directories and copy scripts
+sudo mkdir -p $profiles_dir
+sudo cp -p tools/profile_validate "$bin_dir/buendia-profile-validate"
+sudo cp -p packages/buendia-utils/data/usr/share/buendia/utils.sh "$buendia_dir/utils.sh"
+
+# in the profile_apply script there is a hardcoded reference to utils.sh. sed will replace this a mac friendly url.
+# TODO: review this after system wide path related changes.
+if [[ "$OS" == 'Darwin' ]]; then
+    sed 's/share/local\/opt/g' tools/profile_apply > "$bin_dir/buendia-profile-apply"
```
zestyping: This looks scary... if the word `share` appears anywhere in the script it will get changed.  How about doing this logic in Python in `profile_apply`, rather than using `sed` to edit the Python code?

llvasconcellos: Yes, you are right. I fixed it. Now it looks for /usr/share and replace for /usr/local/opt.

llvasconcellos:
I added a TODO and created an issue #17 to remove the sudo request.

tools/openmrs_setup
```diff
@@ -71,6 +71,33 @@ mysql -uroot -p <<< "
...
+# TODO: remove the need for sudo
+# Create directories and copy scripts
+sudo mkdir -p $profiles_dir
+sudo cp -p tools/profile_validate "$bin_dir/buendia-profile-validate"
+sudo cp -p packages/buendia-utils/data/usr/share/buendia/utils.sh "$buendia_dir/utils.sh"
+
+# In the profile_apply script there is a hardcoded reference to utils.sh. sed will replace this a mac friendly url.
+# TODO: review this after system wide path related changes.
+if [[ "$OS" == 'Darwin' ]]; then
+    sed 's/\/usr\/share/\/usr\/local\/opt/g' tools/profile_apply > "$bin_dir/buendia-profile-apply"
```
viniciusboson: This execution is throwing `tools/openmrs_setup: line 93: /usr/local/bin/buendia-profile-apply: Permission denied`

Refactoring to `sed 's/\/usr\/share/\/usr\/local\/opt/g' tools/profile_apply > sudo "$bin_dir/buendia-profile-apply"`worked fine.

tools/openmrs_setup
```diff
@@ -71,6 +71,33 @@ mysql -uroot -p <<< "
...
+OS=`uname`
+if [[ "$OS" == 'Darwin' ]]; then
+    buendia_dir='/usr/local/opt/buendia'
+    bin_dir='/usr/local/bin'
+fi
+profiles_dir="$buendia_dir/profiles"
+
+# TODO: remove the need for sudo
+# Create directories and copy scripts
+sudo mkdir -p $profiles_dir
```
viniciusboson: You should echo a message informing that the folder xxx will be create.
If not, the user will not even know what password he/she is supposed to fill. 

screen shot 2015-10-23 at 6 45 39 pm

As you can see, the script prompts for the mysql password and then it asks again for a password. What for ?

viniciusboson:
After this  fixing this permission denied, the script worked fine.

screen shot 2015-10-23 at 7 09 48 pm

tools/openmrs_setup
llvasconcellos: Great! Done.

tools/openmrs_setup
llvasconcellos: Well said. Done!

llvasconcellos:
Thank you @viniciusboson. All suggestions accepted and the script is updated.

PR #18, opened 2015-10-27
Fix OSX-specific fatal error in openmrs_setup.

capnfabs:
Previously, `sed` would write to a file named sudo. As of this change, the script writes to the
correct location.

See also: http://stackoverflow.com/q/82256/996592


viniciusboson:
I've tested this PR and it works fine. 
It created the configuration folder and it didn't create the bad file `sudo` anymore.

llvasconcellos:
Great!

PR #20, opened 2015-10-30
Incremental sync for encounters.

capnfabs:
DO NOT MERGE

This change isn't ready to be merged yet; I need to decide on how to branch and merge for this feature.

### DIFF
--- a/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/ProjectBuendiaService.java
+++ b/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/ProjectBuendiaService.java
@@ -11,8 +11,13 @@
 
 package org.projectbuendia.openmrs.api;
 
+import org.openmrs.Encounter;
 import org.openmrs.api.OpenmrsService;
+import org.projectbuendia.openmrs.api.db.ProjectBuendiaDAO;
 import org.springframework.transaction.annotation.Transactional;
+
+import java.util.Date;
+import java.util.List;
 
 /**
  * This service exposes module's core functionality. It is a Spring managed bean which is
@@ -26,5 +31,13 @@ import org.springframework.transaction.annotation.Transactional;
  */
 @Transactional
 public interface ProjectBuendiaService extends OpenmrsService {
-    /* Add service methods here */
+
+    /** Sets the DAO for this service. This is done by DI and Spring. */
+    void setDAO(ProjectBuendiaDAO dao);
+
+    /**
+     * Returns all encounters modified on or after the given {code date}.
+     * @param date if {@code null}, returns all encounters since the beginning of time.
+     */
+    List<Encounter> getEncountersModifiedOnOrAfter(Date date);
 }
--- a/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/ProjectBuendiaDAO.java
+++ b/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/ProjectBuendiaDAO.java
@@ -11,9 +11,14 @@
 
 package org.projectbuendia.openmrs.api.db;
 
+import org.openmrs.Encounter;
 import org.projectbuendia.openmrs.api.ProjectBuendiaService;
+
+import java.util.Date;
+import java.util.List;
 
 /** Database methods for {@link ProjectBuendiaService}. */
 public interface ProjectBuendiaDAO {
-    /* Add DAO methods here */
+    List<Encounter> getEncountersModifiedOnOrAfter(Date date);
+
 }
--- a/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/ProjectBuendiaDAO.java
+++ b/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/ProjectBuendiaDAO.java
@@ -19,6 +19,6 @@ import java.util.List;
 
 /** Database methods for {@link ProjectBuendiaService}. */
 public interface ProjectBuendiaDAO {
-    List<Encounter> getEncountersModifiedOnOrAfter(Date date);
+    List<Encounter> getEncountersCreatedAtOrAfter(Date date);
 
 }
--- a/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAO.java
+++ b/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAO.java
@@ -13,8 +13,14 @@ package org.projectbuendia.openmrs.api.db.hibernate;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.hibernate.Criteria;
 import org.hibernate.SessionFactory;
+import org.hibernate.criterion.Restrictions;
+import org.openmrs.Encounter;
 import org.projectbuendia.openmrs.api.db.ProjectBuendiaDAO;
+
+import java.util.Date;
+import java.util.List;
 
 /** Default implementation of {@link ProjectBuendiaDAO}. */
 public class HibernateProjectBuendiaDAO implements ProjectBuendiaDAO {
@@ -30,5 +36,22 @@ public class HibernateProjectBuendiaDAO implements ProjectBuendiaDAO {
     /** @param sessionFactory the sessionFactory to set */
     public void setSessionFactory(SessionFactory sessionFactory) {
         this.sessionFactory = sessionFactory;
+    }
+
+    @Override
+    public List<Encounter> getEncountersModifiedOnOrAfter(Date date) {
+        // NOTES:
+        // - this code relies on the assumption that observations can't be modified independently of
+        // encounters.
+        // - This doesn't actually return encounters "modified on or after", it's currently
+        // encounters "created on or after". This is ok for our purposes because we don't have
+        // the ability to modify Encounters in Buendia.
+
+        Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Encounter.class);
+        if (date != null) {
+            criteria.add(Restrictions.ge("dateCreated", date));
+        }
+        //noinspection unchecked
+        return criteria.list();
     }
 }
--- a/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAO.java
+++ b/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAO.java
@@ -39,7 +39,7 @@ public class HibernateProjectBuendiaDAO implements ProjectBuendiaDAO {
     }
 
     @Override
-    public List<Encounter> getEncountersModifiedOnOrAfter(Date date) {
+    public List<Encounter> getEncountersCreatedAtOrAfter(Date date) {
         // NOTES:
         // - this code relies on the assumption that observations can't be modified independently of
         // encounters.
--- a/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/impl/ProjectBuendiaServiceImpl.java
+++ b/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/impl/ProjectBuendiaServiceImpl.java
@@ -13,9 +13,14 @@ package org.projectbuendia.openmrs.api.impl;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.openmrs.Encounter;
 import org.openmrs.api.impl.BaseOpenmrsService;
 import org.projectbuendia.openmrs.api.ProjectBuendiaService;
 import org.projectbuendia.openmrs.api.db.ProjectBuendiaDAO;
+
+import javax.annotation.Nullable;
+import java.util.Date;
+import java.util.List;
 
 /** It is a default implementation of {@link ProjectBuendiaService}. */
 public class ProjectBuendiaServiceImpl extends BaseOpenmrsService implements ProjectBuendiaService {
@@ -23,13 +28,13 @@ public class ProjectBuendiaServiceImpl extends BaseOpenmrsService implements Pro
 
     private ProjectBuendiaDAO dao;
 
-    /** @return the dao */
-    public ProjectBuendiaDAO getDao() {
-        return dao;
+    @Override
+    public void setDAO(ProjectBuendiaDAO dao) {
+        this.dao = dao;
     }
 
-    /** @param dao the dao to set */
-    public void setDao(ProjectBuendiaDAO dao) {
-        this.dao = dao;
+    @Override
+    public List<Encounter> getEncountersModifiedOnOrAfter(@Nullable Date date) {
+        return dao.getEncountersModifiedOnOrAfter(date);
     }
 }
--- a/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/impl/ProjectBuendiaServiceImpl.java
+++ b/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/impl/ProjectBuendiaServiceImpl.java
@@ -35,6 +35,6 @@ public class ProjectBuendiaServiceImpl extends BaseOpenmrsService implements Pro
 
     @Override
     public List<Encounter> getEncountersModifiedOnOrAfter(@Nullable Date date) {
-        return dao.getEncountersModifiedOnOrAfter(date);
+        return dao.getEncountersCreatedAtOrAfter(date);
     }
 }
--- a/openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
+++ b/openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
@@ -157,7 +157,7 @@ public class EncounterResource
      *   <li>"timestamp": the encounter datetime in RFC 3339 === ISO 8601 format
      *   <li>"uuid": the encounter's UUID
      *   <li>"observations": {@link SimpleObject} that maps concept UUIDs to values
-     *   <li>"order_uuids": unique identifiers of orders related to this encounter.
+     *   <li>"order_uuids": unique identifiers of orders executed as part of this encounter.
      * </ul>
      * @param context      unused.
      * @param snapshotTime unused.
--- a/openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
+++ b/openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
@@ -26,8 +26,10 @@ import org.openmrs.module.webservices.rest.web.resource.api.Creatable;
 import org.openmrs.module.webservices.rest.web.response.ResponseException;
 import org.openmrs.projectbuendia.Utils;
 import org.openmrs.projectbuendia.VisitObsValue;
+import org.projectbuendia.openmrs.api.ProjectBuendiaService;
 import org.projectbuendia.openmrs.webservices.rest.RestController;
 
+import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
@@ -41,27 +43,31 @@ import java.util.List;
 @Resource(name = RestController.REST_VERSION_1_AND_NAMESPACE + "/encounters",
     supportedClass = Patient.class, supportedOpenmrsVersions = "1.10.*,1.11.*")
 public class EncounterResource
-    extends AbstractReadOnlyResource<Patient> implements Creatable {
+    extends AbstractReadOnlyResource<Encounter> implements Creatable {
     private final PatientService patientService;
-    private final EncounterService encounterService;
+    private final ProjectBuendiaService buendiaService;
 
     public EncounterResource() {
-        super("patient", Representation.DEFAULT);
+        super("encounter", Representation.DEFAULT);
         patientService = Context.getPatientService();
-        encounterService = Context.getEncounterService();
+        buendiaService = Context.getService(ProjectBuendiaService.class);
     }
 
     /**
-     * Returns all patients.  The retrieved records will be filled in with
-     * each patient's encounter and observation data by
-     * {@link #populateJsonProperties(Patient, RequestContext, SimpleObject, long)}
-     * on its way to becoming JSON that is sent to the client.
-     * @param context      unused here; see populateJsonProperties() for details
+     * Returns all encounters observed after the "since" parameter, or since the beginning of time
+     * if the "since" parameter is not set.
+     * @param context      used to obtain the "since" parameter for incremental sync.
      * @param snapshotTime unused here; see populateJsonProperties() for details
      * @see AbstractReadOnlyResource#search(RequestContext)
      */
-    @Override public List<Patient> searchImpl(RequestContext context, long snapshotTime) {
-        return patientService.getAllPatients();
+    @Override public List<Encounter> searchImpl(RequestContext context, long snapshotTime) {
+        Date syncFrom;
+        try {
+            syncFrom = RequestUtil.getSyncFromDate(context);
+        } catch (ParseException e) {
+            throw new IllegalArgumentException("Date Format invalid, expected ISO 8601");
+        }
+        return buendiaService.getEncountersModifiedOnOrAfter(syncFrom);
     }
 
     /**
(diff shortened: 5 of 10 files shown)


openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAO.java
```diff
@@ -31,4 +37,21 @@ public SessionFactory getSessionFactory() {
     public void setSessionFactory(SessionFactory sessionFactory) {
         this.sessionFactory = sessionFactory;
     }
+
+    @Override
+    public List<Encounter> getEncountersModifiedOnOrAfter(Date date) {
```
zestyping: This is really cool. Thanks for showing how to do this.

openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAO.java
```diff
@@ -31,4 +37,21 @@ public SessionFactory getSessionFactory() {
...
+        // NOTES:
+        // - this code relies on the assumption that observations can't be modified independently of
+        // encounters.
+        // - This doesn't actually return encounters "modified on or after", it's currently
+        // encounters "created on or after". This is ok for our purposes because we don't have
```
zestyping: Any reason not to name the method ...CreatedAtOrAfter?

openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/ProjectBuendiaDAO.java
```diff
@@ -11,9 +11,14 @@
...
+import org.openmrs.Encounter;
 import org.projectbuendia.openmrs.api.ProjectBuendiaService;
 
+import java.util.Date;
+import java.util.List;
+
 /** Database methods for {@link ProjectBuendiaService}. */
 public interface ProjectBuendiaDAO {
-    /* Add DAO methods here */
+    List<Encounter> getEncountersModifiedOnOrAfter(Date date);
```
zestyping: This is a tiny thing, but how about "at" instead of "on" (because "on" sustains the misunderstanding that Date is a date, but it isn't, it's a timestamp).

openmrs/api/src/main/java/org/projectbuendia/openmrs/api/impl/ProjectBuendiaServiceImpl.java
```diff
@@ -13,23 +13,28 @@
...
+    @Override
+    public void setDAO(ProjectBuendiaDAO dao) {
+        this.dao = dao;
     }
 
-    /** @param dao the dao to set */
-    public void setDao(ProjectBuendiaDAO dao) {
-        this.dao = dao;
+    @Override
+    public List<Encounter> getEncountersModifiedOnOrAfter(@Nullable Date date) {
```
zestyping: Just wondering -- is there any way to make this return an iterator, so that if we want to return just the first page of 100 results we can do that without fetching the entire list?  Or are we stuck with the whole list because Criteria offers no other option?

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
```diff
@@ -39,43 +40,34 @@
  * point in time (referred to as the "snapshot time").
  * @see AbstractReadOnlyResource
  */
-// TODO: Merge with PatientResource and let clients use a query parameter to
-// indicate whether encounter data should be included with each returned patient.
 @Resource(name = RestController.REST_VERSION_1_AND_NAMESPACE + "/encounters",
     supportedClass = Patient.class, supportedOpenmrsVersions = "1.10.*,1.11.*")
 public class EncounterResource
-    extends AbstractReadOnlyResource<Patient> implements Creatable {
+    extends AbstractReadOnlyResource<Encounter> implements Creatable {
```
zestyping: Ahhhh, much better.:)

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
```diff
@@ -369,38 +367,18 @@ private static void setLocation(Patient patient, String locationUuid) {
...
         String patientId = requestContext.getParameter("id");  // single patient by ID; overrides
-        // "q"
         if (patientId != null) {
             List<PatientIdentifierType> idTypes = new ArrayList<>();
             idTypes.add(DbUtil.getMsfIdentifierType());
             patients = patientService.getPatients(null, patientId, idTypes, true);
         } else {
-            String nameQuery = requestContext.getParameter("q");
-            patients = filterPatients(nameQuery, patientService.getAllPatients());
+            patients = patientService.getAllPatients();
```
zestyping: Any particular reason why this deletion is part of this PR?  It's true that the client doesn't use this at the moment, just wondering if it could be a separate change.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/RequestUtil.java
```diff
@@ -0,0 +1,40 @@
...
+     * @return the instant that data should be provided from (inclusive), or {@code null} if the
+     * "since" parameter was not present in the request.
+     */
+    @Nullable
+    public static Date getSyncFromDate(RequestContext context) throws ParseException {
+        String param = context.getParameter("since");
+        if (param == null) {
+            return null;
+        }
+        return Utils.fromIso8601(param);
```
zestyping: I'm hoping to get all representations of time on the wire to use millis since epoch only -- it reduces redundant format/parse steps and potential bugs around time zone conversions despite the time zone information being irrelevant to the semantics of the parameter. In my fantasy world all our datastores would only ever store millis as well, and we would eliminate all use of formatted dated and times except only for display to the user.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
```diff
@@ -145,17 +137,35 @@ public Object createInner(SimpleObject post, RequestContext context) throws Resp
...
+        return null;
+    }
 
+    /**
+     * Populates observation and order data for the given encounter, including:
+     * <ul>
+     *   <li>"timestamp": the encounter datetime in RFC 3339 === ISO 8601 format
+     *   <li>"uuid": the encounter's UUID
+     *   <li>"observations": {@link SimpleObject} that maps concept UUIDs to values
+     *   <li>"order_uuids": unique identifiers of orders related to this encounter.
```
zestyping: I named this poorly.  It should be clear that these are orders _executed_ during this encounter.

Orders are also _created_ during encounters, which means that if there is also a field for new orders here, we'll also get incremental order sync!  Some care needs to be taken to handle order editing/cancellation properly as well as order UUID substitution -- see OrderResource for details on this and the rationale.  The "edit stored as add" convention used there should fit in nicely with incremental sync though.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
```diff
@@ -39,43 +40,34 @@
...
-            }
-            List<Patient> patients = new ArrayList<>();
-            patients.add(patient);
-            return patients;
-        } else {
-            return patientService.getAllPatients();
+    @Override public List<Encounter> searchImpl(RequestContext context, long snapshotTime) {
+        Date syncFrom;
+        try {
+            syncFrom = RequestUtil.getSyncFromDate(context);
```
zestyping: Potential for a subtle bug:

Suppose our page size is 100 entries and there are 101 encounters with the same timestamp. Then there is no way the client can see the last encounter.

Unlikely to occur due to user entry but still, I could totally see it happening as a result of data import or initializing the database with test data (e.g. 1000 dummy patients).

The only solution I've thought of so far is to make the bookmark consist of a timestamp + encounter UUID, and sort by both timestamp and secondarily by UUID when querying the db.  (Hm, I like the name "bookmark"... seems evocative and appropriate for this scenario, much more appropriate than the "favourite URL" meaning of "bookmark"...)

capnfabs: Good catch. That will be problematic. I was going to keep pagination and the `since` timestamp separate (see discussion on the doc), but we still have a problem under that scenario:
- Page size 100
- 101 encounters, all with the same timestamp
?since=X&page=0 returns 100 unique results
?since=X&page=1 is likely to return one of the previous 100 results instead of the final missing encounter, because ORDER BY makes no guarantees that ordering is stable for non-unique rows.

The solution is indeed to ORDER BY timestamp, uuid. Note that, however, if the timestamp isn't used as a pagination mechanism, we only need to ORDER BY uuid, because we don't care about the order that the results are retrieved in, just that they're all retrieved.

capnfabs: Also, "bookmark" is a great name for this concept. Regarding your point about avoiding confusion between client and server timestamps, I was thinking about making the since= parameter a token that's opaque to the client app so there's no room for confusion. I don't think it's entirely necessary right now, but as a "nice to have" we could do something like base64-encode the server-provided timestamps and label them as "sync tokens", which would prevent the client from misusing them accidentally.

zestyping: But uuids don't necessarily increase over time, so if we only sort by uuid
then the client is likely to refetch half the entries every time instead of
just the new ones... am I misunderstanding what you meant?

capnfabs: Maybe. My intended design is to:

- Filter by the `since=` parameter, which is a timestamp previously supplied by the server.
- Paginate using a `page=` parameter, which is independent from modification time
- Sort the data by UUID, to ensure stable ordering.

So the resultant SQL query approximately looks like:

    WHERE date_modified >= [since parameter] ORDER BY uuid LIMIT [page_num*page_size],[(page_num + 1)*page_size]

So it doesn't matter that UUIDs aren't monotonically increasing, just that they're unique, which provides stable ordering such that the LIMIT clause provides consistent results across consecutive queries.

Does that make more sense?

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
capnfabs: There isn't really a reason why this deletion is part of the PR, just that I hadn't got my head around how to do branching for this feature yet. I'll separate this commit out and send a seperate PR.

FWIW, refactoring sync is going to result in a number of breaking protocol changes. I was thinking I'd branch `dev` to `feature/sync-overhaul` in both the `projectbuendia/buendia` `projectbuendia/client` repos and then send PRs to `feature/sync-overhaul`. This allows for discussion on chunks of code along the way, whilst keeping work separate so as not to break other developers. Then at the end, we do a big merge back into `projectbuendia/buendia/dev`. Does that approach sound sensible to you?

openmrs/api/src/main/java/org/projectbuendia/openmrs/api/impl/ProjectBuendiaServiceImpl.java
capnfabs: It looks like Hibernate has a `ScrollableResults` class that we can use instead. We could make it return an iterator to support pagination further up the stack - that's probably a better design than passing pagination info down the stack - but I'll do this as part of the pagination changes instead of now.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
zestyping: Yes, maintaining a branch in each repo with the same name sounds pretty reasonable.  (Try to name it something more specific than sync-overhaul if you can.)

The client repo is a submodule of the buendia repo; the intention there was to use the submodule pointer to track coordinated changes between client and server (i.e. when you check out the entire buendia repo recursively you should get a client and server that interoperate).  In practice this means you'd merge the client feature branch to dev first; then in the server feature branch, update the client submodule to point at the newly merged client; then merge the server feature branch to dev.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
zestyping: I see what you mean.  In this intended design, what happens when two different clients submit encounters with the same timestamp?  The UUIDs aren't guaranteed to arrive at the server in increasing order.

openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAO.java
capnfabs: Good call. Done.

openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/ProjectBuendiaDAO.java
capnfabs: Done.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
capnfabs: Awesome, that's a great design. Thanks for the info! I've created feature/sync-overhaul on the both the `buendia` and the `client` repos, I'll be sending PRs to those from now on.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/RequestUtil.java
capnfabs: The parse/format time is something that I hadn't thought about. I suspect that the time it takes to do that is dwarfed by the rest of the request, though (JSON, I'm looking at you!).

I like the use of ISO-8601 because it makes debugging the REST API heaps easier, especially for incremental sync, where correct timestamps are critical. Initially, I'd designed the protocol to use millis since epoch in UTC, but I made the following programming errors:
- At one point, I was supplying epoch time in seconds, instead of millis, and so more results than I was expecting were being returned. This would not be a problem using a human-readable format.
- I was never really 100% sure what timezone I was operating in. The fact that ISO-8601 makes this explicit is awesome.

I think I'd like to retain ISO-8601 for the API for ease of debugging at minimal performance cost, but to make the protocol explicit in saying "only the UTC timezone is accepted".

Does that sound reasonable to you?

Also, FWIW, it seems like readable / standards-based date/time formats are much more common for REST APIs than millis (e.g. Facebook uses ISO 8601, Twitter uses RFC5322)

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
capnfabs: I've amended the comment to state that this is orders executed.
Thanks for the information re incremental order sync. It sounds complex and is certainly beyond what I understand at the moment, so I'll address that in a future PR.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
capnfabs: Oh, I understand what you're saying now! Sorry that it's taken a while.

Yes, the way I've described it above has a problem if:
- Tablet A is syncing data, and has 7 out of 10 pages
- Tablet B uploads a new encounter, with a UUID that falls within the first 70% of UUIDs on the system.

using `ORDER BY date, uuid` works around this, but will still result in the number of pages changing mid-sync. Which is ok using a `more` mechanism, and weird using a `num_pages` mechanism. Your idea of using [timestamp + uuid] as the bookmark doesn't suffer from this problem either.

There is also still an (extremely unlikely) edge case where this solution doesn't work.
- Records are being bulk added, so multiple records are being added with the same timestamp.
- Halfway through the bulk add, the client requests a page.
- More records are bulk added with the same timestamp
- The client requests another page, only to find that the pages are different.

capnfabs: So, it seems that we can never be 100% certain that:
- we won't fetch the same data twice
- we will fetch all data on a single sync.

That's probably ok, though, because:
- Our data model is ok with records being eventually consistent. See the [sync doc](//j.mp/buendia-sync) under 'Eventual Consistency'
- Because we're using `INSERT OR REPLACE` for everything on the client side, it's ok if we fetch the same record multiple times.

In terms of guaranteeing that everything still works under the edge case, the client should save the server timestamp associated with the _first_ page, and then re-fetch from there on the next incremental sync. It's ok if the records added during sync are fetched a non-deterministic number of times because we're using `INSERT OR REPLACE`, as mentioned above.

Does that make sense?

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
capnfabs: Also, created pull #22 for the cleanup commits.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/RequestUtil.java
zestyping: Perhaps it's just another difference in taste?  I hate/worry about time zone issues so much (and have been bitten by costly time zone bugs so many times) that using a format in which it is _impossible_ to express anything other than UTC is worth a heckuva lot to me.  To me avoiding this class of bugs greatly outweighs the (subjectively) small cost of converting millis to y/m/d/h/m/s when I'm debugging.

It's also nice that every database and programming language knows how to store and handle integers, but y/m/d/h/m/s structs are handled in lots of different ways and many systems have no primitive.  For example, 2016-02-29T00:00:00Z is parseable but 2015-02-29T00:00:00Z is not, and whether 2015-12-19T00:00:60Z is parseable is implementation-dependent.  If one can look at data on the wire and be unable to tell whether it's well-formed, that feels pretty tragic.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
zestyping: Eh, the submodule features in git are somewhat confusing to use, so I don't exactly love this scheme, but it's the best I've been able to think of so far.  There's room for improvement I'm sure.

capnfabs:
I'm going to close this pull request and open a new one targeting the `feature/incremental-sync` branch, as mentioned earlier in the PR discussion.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
zestyping: Agree it's okay to fetch the same record multiple times, or a non-deterministic number of times. Also agree that your edge case exists.  I'm not seeing how saving the timestamp associated with the first page solves the edge case though.  If the client keeps going back to the first timestamp, how does it ever advance to fetching new data?

The number of pages can change mid-sync even without a UUID collision; in fact I'm realizing that having the number of pages change mid-sync will be a common occurrence in practice.  For example, the page size is 100, the client has all but the last 99 encounters; while the client is syncing those 99, 2 more encounters are submitted.  Given that we can't reliably tell the client how many pages there are, why report num_pages at all?  Also if the client is interrupted before it finishes fetching all the pages, what happens then—is "I tried to fetch n pages but only finished fetching k" a special state that the client has to remember, different from the regular "let's ask the server for more records" state?

All this feels to me like it pushes us toward a design based on a single sync token (probably (timestamp, UUID) or equivalent), both for better simplicity and better correctness.  I think the Google APIs got it right.  What do you think?

capnfabs: Which Google APIs are you talking about?

capnfabs: Ok. I spent another few hours thinking about this this morning; I'm convinced about using timestamp + UUID now :)

A few observations:
- The biggest strength of [timestamp + UUID] is that it means that every HTTP request + apply to local DB results in a determinate state - we don't ever need to rollback transactions if a sync only partially completes. That's extremely powerful, and I think worth having.
- I think neither bookmarking method ([timestamp + UUID], [timestamp + page number]) solves the problem where we could miss records added at the same time as the final page is synced. This is something that we have to address by choosing the bookmark returned for the last page of results. I have a proposal for this, I'll add it to the sync doc.
- The question of whether to report `num_pages` at all is slightly orthogonal to the choice of bookmark. It might still be worth providing an estimated number of pages to the client in each request so that they can indicate sync progress, but we can do that in parallel/as an enhancement now.

One meta-observation:
- I think it probably would've been much quicker if we had've hopped online and chatted about this in person rather than debating it on the thread :) Next time let's chat about this in person.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/RequestUtil.java
capnfabs: Millis can still suffer from timezone problems though - you're just hiding it by rolling it into the actual timestamp. I'm going to keep this as ISO 8601; it's easier to debug (which is a huge plus for me given how much request debugging I'll be doing in the coming weeks :wink: ), and seeing as it's treated as a token by the client (i.e. no need to parse it), we don't have to worry as much about compatibility - we're only working with a single implementation.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
zestyping: The Calendar API was the one I had in mind (search for "nextSyncToken").

Sorry, I thought I had seen this as a common technique in a few APIs, but
now I can only find it in the Calendar API!

zestyping: Agreed on all points! :)

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/RequestUtil.java
zestyping: Okay.  This is definitely your call (and honestly I can see your side of it
too).

openmrs/api/src/main/java/org/projectbuendia/openmrs/api/ProjectBuendiaService.java
```diff
@@ -26,5 +31,13 @@
...
-    /* Add service methods here */
+
+    /** Sets the DAO for this service. This is done by DI and Spring. */
+    void setDAO(ProjectBuendiaDAO dao);
+
+    /**
+     * Returns all encounters modified on or after the given {code date}.
+     * @param date if {@code null}, returns all encounters since the beginning of time.
+     */
+    List<Encounter> getEncountersModifiedOnOrAfter(Date date);
```
dancunningham: Might we also want to constrain this to a) a single location and/or b) a single patient?

openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAO.java
```diff
@@ -31,4 +37,21 @@ public SessionFactory getSessionFactory() {
...
     }
+
+    @Override
+    public List<Encounter> getEncountersCreatedAtOrAfter(Date date) {
+        // NOTES:
+        // - this code relies on the assumption that observations can't be modified independently of
+        // encounters.
+        // - This doesn't actually return encounters "modified on or after", it's currently
+        // encounters "created on or after". This is ok for our purposes because we don't have
+        // the ability to modify Encounters in Buendia.
```
dancunningham: I agree that should be fine. Encounters or observations should never be modified. To correct an observation, we would void the old observation and add a new observation with the new value.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/EncounterResource.java
```diff
@@ -39,43 +40,34 @@
...
-            Patient patient = patientService.getPatientByUuid(patientUuid);
-            if (patient == null) {
-                throw new ObjectNotFoundException();
-            }
-            List<Patient> patients = new ArrayList<>();
-            patients.add(patient);
-            return patients;
-        } else {
-            return patientService.getAllPatients();
+    @Override public List<Encounter> searchImpl(RequestContext context, long snapshotTime) {
```
dancunningham: What does "searchImpl" mean?

capnfabs: It's just the way this code was structured initially. The OpenMRS REST API is quite prescriptive, so it calls the following methods under the following circumstances:

- `getAll()` - if `/encounters` is requested with no params
- `search()` - if `/encounters` is requested with params, e.g. `/encounters?thingy=foo`

At some time in the past, someone on our team wrote this `AbstractReadOnlyResource` that adds a bunch of convenience functionality to these methods. It's still up to subclasses to provide data, however, so `AbstractReadOnlyResource` has abstract `getAllImpl`, `searchImpl` methods that subclasses are supposed to override.

openmrs/api/src/main/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAO.java
capnfabs: Actually, this assumption breaks with @G1DR4's change to void observations - once an observation can be voided, an encounter can logically be modified (because encounters have multiple observations). I'll need to adjust this logic when that change goes through.

openmrs/api/src/main/java/org/projectbuendia/openmrs/api/ProjectBuendiaService.java
capnfabs: Maybe! But we can do that later, and possibly at a higher level of the API than the DB layer.

PR #22, opened 2015-11-02
Cleanup / Remove unused query params

capnfabs:
Neither of these parameters are used by the client, so they're safe to remove. I'm interested in removing them because they're additional overhead that would otherwise need to be maintained as part of the sync refactor.


capnfabs:
ok to merge?

viniciusboson:
I think it's fine too.

PR #24, opened 2015-11-03
Cleanup some API method names for better consistency

capnfabs:
In response to comments on #20, I renamed some API methods from `getEncountersModifiedOnOrAfter` to `getEncountersCreatedAtOrAfter`. I missed a couple of places though.


viniciusboson:
It's perfect to me.

PR #25, opened 2015-11-04
Allow concept types to be changed by profile_apply script

capnfabs:
DO NOT MERGE

This what I had to do to apply dev.projectbuendia.org's current profile locally.

### DIFF
--- a/tools/profile_apply
+++ b/tools/profile_apply
@@ -235,10 +235,10 @@ def apply(tabs):
         """Creates or updates a concept with the given name, type, and class."""
         datatype_id = db.get('datatype_id', concept_id=concept_id)
         # Changing the datatype of an existing concept is illegal in OpenMRS.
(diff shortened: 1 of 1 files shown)


llvasconcellos:
I'm lost regarding to profile_apply. I can't be of any help by now.

capnfabs:
No worries. Like what @viniciusboson was saying in standup, I'm not sure that this is the right thing to do - probably better to work out how to use these concepts correctly rather than just do this :)

I'm just trying to document what's required so that if people get stuck in the future they've got options.

viniciusboson:
Just to be documented as well, [here is more comments](https://github.com/projectbuendia/client/issues/53#issuecomment-152531404) of that, on #53.

capnfabs:
Nice! I'd missed that before. Thanks @viniciusboson

PR #27, opened 2015-11-05
Change the dev-specific SQL script to use Bokoro locations.

capnfabs:
All the tests are written assuming Bokoro locations, so the dev setup script should use them too,
so that tests can be run locally.

This is a straight copy-paste of the site-msf-chad-bokoro.sql file.


viniciusboson:
These changes look ok to me ...

llvasconcellos:
Wow, that's great!

PR #28, opened 2015-11-05
Rename Gender --> Sex as per TODO

capnfabs:
Corresponding client change: https://github.com/projectbuendia/client/pull/126

### DIFF
--- a/openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
+++ b/openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
@@ -65,7 +65,7 @@ import java.util.Map;
  *   "uuid": "e5e755d4-f646-45b6-b9bc-20410e97c87c", // assigned by OpenMRS, not required for
  *   creation
  *   "id": "567", // required unique id specified by user
- *   "gender": "F", // required as "M" or "F", unfortunately
+ *   "sex": "F", // required as "M" or "F", unfortunately
  *   "birthdate": "1990-02-17", // required, but can be estimated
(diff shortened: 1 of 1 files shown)


openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
```diff
@@ -463,10 +460,6 @@ protected void applyEdits(Patient patient, SimpleObject edits) {
                     patient.setBirthdate(Utils.parseLocalDate((String) entry.getValue(), BIRTHDATE));
                     changedPatient = true;
                     break;
-                case GENDER:
-                    patient.setGender(normalizeSex((String) entry.getValue()));
-                    changedPatient = true;
-                    break;
```
llvasconcellos: We don't need this anymore?

llvasconcellos:
If we don't need GENDER, this should be good to go.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
capnfabs: Not once the corresponding client change is checked in :) I'm taking the opportunity whilst doing the sync refactor to clean up TODOs that would result in breaking the protocol, because the sync refactor breaks the protocol anyway.

capnfabs:
Note that this is a pull request against the `feature/incremental-sync` branch, we're I'm making all the changes for incremental syncing. I'll merge it back into `dev` once all the protocol-breaking changes are done, to minimise disruption in the meantime.

PR #29, opened 2015-11-09
Add incremental sync support for patients.

capnfabs:


llvasconcellos:
No comments on the code. It looks good to me. 
Two things:
* How can I test it?
* Are you going to write tests for it?

capnfabs:
Thanks Leonardo!

- Test it by spinning up a local server built from the equivalent branch and running the standard client tests against the local server
- I'm not going to write tests for it yet; for now, all the other tests depend on sync working so we should see widespread failure if it doesn't :) Longer term, I'd like to refactor the sync code and enough of the network stack to be able to unit-test sync instead of relying upon testing with the integration tests. But that's quite a bit of work, and probably beyond the scope of this sprint.

PR #30, opened 2015-11-10
Provide an endpoint for syncing observations directly instead of sync…

capnfabs:
…ing encounters.

Encounter sync worked well whilst we could guarantee that encounters and their associated
observations could never be modified, but that will change soon, as we introduce the ability to void
patients.

Note that when an individual observation is voided in OpenMRS, it makes no change to the encounter
record.

There are two solutions that allow for observations to be voided, then:
- Change the encounter modification detection logic such that it queries for observations that have
  been voided or modified. Then, if a modified encounter is detected, send the entire encounter to
  the client. The client then has to drop all existing records for the encounter, and re-add the
  ones present in the response.
- Change so that we sync observations instead of syncing encounters. This makes the modification
  detection logic much more straightforward and standard, and makes the implementation on the client
  side more consistent as well.

Note that we have some weirdness now in that the client always writes encounters, and reads
observations as a result of those encounters. I don't like the asymmetry there, but we already had
that anyway in the way that we store data and the way that we display it.

### DIFF
--- a/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/ProjectBuendiaService.java
+++ b/openmrs/api/src/main/java/org/projectbuendia/openmrs/api/ProjectBuendiaService.java
@@ -11,7 +11,7 @@
 
 package org.projectbuendia.openmrs.api;
 
-import org.openmrs.Encounter;
+import org.openmrs.Obs;
 import org.openmrs.api.OpenmrsService;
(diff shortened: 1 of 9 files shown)


capnfabs:
Assigning @G1DR4 for review because it's relevant to his work :)

G1DR4:
Fabian made my day with this pull. Good to go.

PR #31, opened 2015-11-11
True incremental sync for patients and observations.

capnfabs:
Note: this is a *breaking protocol change*.
See corresponding PR in the Android client: https://github.com/projectbuendia/client/pull/140

For the high-level design, see http://j.mp/buendia-sync. Note that pagination from that document hasn't been implemented yet - just incremental sync.


capnfabs:
FYI, this code has all been reviewed, and is ready to merge as soon as we're willing to deal with the fact that it's a breaking protocol change.

capnfabs:
🚀

PR #33, opened 2015-11-12
Make the buendia-update script actually error if one of its steps fails.

capnfabs:
We had a problem where Jenkins wasn't pushing new builds to dev.projectbuendia.org, and we had no
idea because the buendia-update script was squelching errors.

This change removes some of the squelching such that update failures cause Jenkins to actually show
failures.

### DIFF
--- a/packages/buendia-update/data/usr/bin/buendia-update
+++ b/packages/buendia-update/data/usr/bin/buendia-update
@@ -10,7 +10,9 @@
 # OR CONDITIONS OF ANY KIND, either express or implied.  See the License for
 # specific language governing permissions and limitations under the License.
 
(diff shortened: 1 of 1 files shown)


zestyping:
It looks like what this will do is:
  - If any package is partially installed, no other packages will ever be updated
  - If any package has a problem during update, the server won't reboot even if required by another package

Are you sure you want that?

capnfabs:
No, that's not what I wanted. I've revised the script so that a partially-installed package won't cause the script to exit. Basically now, the whole thing executes every time, just the exit code is non-zero if the install failed.

dancunningham:
Looks good to me!

PR #37, opened 2015-11-16
Fix the datatype of `buendia-concept-order_executed`.

capnfabs:
Previously, the datatype was `N/A`, but we're storing a numeric value in there.

This is an inconsistency in the data model that was causing validation errors for certain operations
on observations.

Note that this change has no impact on the user experience, or in the way the system works - it just
resolves a problem that I hit when tinkering.

I'm not sure if this is the best way to resolve this inconsistency - other valid approaches would
be to
- Simply not store a value with the observation, or
- Change the type of the concept to `Boolean`

It seems possible to me that it could be useful to keep this numeric to indicate (for example)
a double or half-dose.


capnfabs:
Actually, this doesn't work for some reason I didn't get around to debugging. Closing.

PR #38, opened 2015-11-20
Incremental Sync for Orders.

capnfabs:
Probably best to review this commit-by-commit; the first two commits are cleanup only and the third is the actual incremental sync logic.

### DIFF
--- a/openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
+++ b/openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
@@ -36,17 +36,13 @@ import org.openmrs.module.webservices.rest.web.resource.api.Searchable;
 import org.openmrs.module.webservices.rest.web.resource.api.Updatable;
 import org.openmrs.module.webservices.rest.web.response.ObjectNotFoundException;
 import org.openmrs.module.webservices.rest.web.response.ResponseException;
+import org.projectbuendia.openmrs.api.ProjectBuendiaService;
 import org.projectbuendia.openmrs.webservices.rest.RestController;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
-import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
-import java.util.Map;
-import java.util.Set;
 
 /**
  * Rest API for orders.
@@ -66,6 +62,7 @@ import java.util.Set;
  * {
  *   "uuid": "e5e755d4-f646-45b6-b9bc-20410e97c87c", // assigned by OpenMRS, not required for
  *   creation
+ *   "voided": false, // If true, fields other than UUID are not guaranteed to be set.
  *   "instructions": "Paracetamol 2 tablets 3x/day",
  *   "start_millis": 1438711253000,
  *   "stop_millis": 1438714253000  // optionally present
@@ -99,12 +96,14 @@ public class OrderResource implements Listable, Searchable, Retrievable, Creatab
     private final OrderService orderService;
     private final ProviderService providerService;
     private final EncounterService encounterService;
+    private final ProjectBuendiaService buendiaService;
 
     public OrderResource() {
         patientService = Context.getPatientService();
         orderService = Context.getOrderService();
         providerService = Context.getProviderService();
         encounterService = Context.getEncounterService();
+        buendiaService = Context.getService(ProjectBuendiaService.class);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
```diff
@@ -91,74 +87,57 @@
...
 )
 public class OrderResource implements Listable, Searchable, Retrievable, Creatable, Updatable {
-    static final User CREATOR = new User(1);  // fake value
-    static final RequestLogger logger = RequestLogger.LOGGER;
-    static Log log = LogFactory.getLog(OrderResource.class);
-    static final String FREE_TEXT_ORDER_UUID = "buendia-concept-free_text_order";
+    private static final User CREATOR = new User(1);  // fake value
+    private static final RequestLogger logger = RequestLogger.LOGGER;
+    private static Log log = LogFactory.getLog(OrderResource.class);
+    private static final String FREE_TEXT_ORDER_UUID = "buendia-concept-free_text_order";
```
llvasconcellos: I noticed lots of missing access modifiers on the server code (profile manager). I also added those, as you can see here:
https://github.com/projectbuendia/buendia/pull/14/files

I also do prefer to specify it even if it is protected.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
```diff
@@ -91,74 +87,57 @@
...
+    private final EncounterService encounterService;
+    private final ProjectBuendiaService buendiaService;
 
     public OrderResource() {
         patientService = Context.getPatientService();
         orderService = Context.getOrderService();
         providerService = Context.getProviderService();
-        conceptService = Context.getConceptService();
         encounterService = Context.getEncounterService();
+        buendiaService = Context.getService(ProjectBuendiaService.class);
```
llvasconcellos: Teach me a bit of java, please? When you do this you get ProjectBuendiaServiceImpl? You don't need to type cast it?

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
```diff
@@ -91,74 +87,57 @@
...
-    }
+    private SimpleObject handleSync(RequestContext context) throws ResponseException {
+        Date date = RequestUtil.mustParseSyncFromDate(context);
 
-    SimpleObject getSimpleObjectWithResults(Collection<Order> orders) {
-        List<SimpleObject> jsonResults = new ArrayList<>();
+        // Set the next sync from time to be one second in the past because there's issues
+        // due to OpenMRS rounding truncating insertion times. See
+        // https://github.com/projectbuendia/client/pull/90
+        Date nextSyncFrom = new Date(System.currentTimeMillis() - 1000);
```
llvasconcellos: Did you run the Tests? Did it pass?

llvasconcellos:
Looks good to me.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
capnfabs: Sure. This uses Java generics to avoid the typecast. This method is actually defined as:
```java
public static <T> T getService(Class<? extends T> cls);
```
So when you pass `ProjectBuendiaService.class`, as a parameter, the java compiler is able to infer the value of `T`, and thus the method returns a `ProjectBuendiaService`.

The official [Java generics tutorial](https://docs.oracle.com/javase/tutorial/java/generics/index.html) might be helpful.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
capnfabs: Yep! That's how I worked out the tests were failing in https://github.com/projectbuendia/client/pull/153 :)

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
llvasconcellos: :+1:

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
llvasconcellos: :+1:

PR #39, opened 2015-11-24
Implement pagination for patients.

capnfabs:
Note that it's not actually in use yet - we still need to pass a full sync token to and from the
client.


viniciusboson:
Nice work, @capnfabs. And it's really well documented. 

Just one question. How are you used to work with Liquibase ? I prefere to keep one file for each change set. If so, what do think about given a more suggestive name for liquibase file ? I'd rename it to `20151124224143_add_person_date_updated_index.xml` for example, and i would put them into a specific folder, so that it would be both more organized and readable as long as the changes are increasing.

capnfabs:
Thanks @viniciusboson! This is the first time I've worked with Liquibase actually. I found some other advice around the internet suggesting the same thing as you, but it looks like OpenMRS modules [only support having a single file named liquibase.xml](https://wiki.openmrs.org/display/docs/Module+liquibase+File).

viniciusboson:
Oh, too bad !!!

capnfabs:
I've got a new approach that is much easier to test and probably a little bit less hacky. PR incoming.

PR #40, opened 2015-11-27
Implement Patient Sync pagination in the DAO

capnfabs:
This is a pretty different approach to the old PR #39.

The old approach:
- created a new DB column on the `person` table, 
- updated that column with a SQL trigger,
- fought against Hibernate to be able to use that column in WHERE and ORDER BY clauses.

The new approach:
- creates a new DB _table_, that's keyed off the `person` table
- updates records in that table with a SQL trigger (similarly to the old approach)
- Accesses those records using Hibernate - Hibernate treats the sync data as actual objects now, which makes them significantly easier to work with.

I rewrote this code using the new approach because it seemed like the old approach wasn't testable. For tests, OpenMRS uses an in-memory database, which has its schema defined by Hibernate's ideas about what the tables should look like on test startup. It is actually possible to test everything under the old method, it just requires making changes to the in-memory DB at test startup to add the appropriate columns.

It seems to me that there's tradeoffs in both approaches - the old approach is heaps cleaner in terms of database representation and updating sync timestamps, whereas the new approach is much cleaner in actually accessing the data. I could get rid of most of the comment on `HibernateProjectBuendiaDAO` that explained the big list of workarounds in use. I don't have strong opinions as to which is better though. I suspect that this one will be more resilient against changes to OpenMRS in future versions, but possibly less performant, and with a less clean DB schema.

As an aside, this new direction would allow us to even consider creating the update records in the Application layer, instead of the DB layer, which could provide better db-independance in the future.

### DIFF
--- a/openmrs/api/src/test/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAOTest.java
+++ b/openmrs/api/src/test/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAOTest.java
@@ -112,8 +112,8 @@ public class HibernateProjectBuendiaDAOTest extends BaseModuleContextSensitiveTe
     /**
      * {@link BaseModuleContextSensitiveTest} does this initialization, but also pre-loads the
      * database with a bunch of patient records. We don't want to load those patient records,
-     * because we'd then have to augment them with `buendia_patient_sync_map` records, which would couple
-     * our test integrity to the records in OpenMRS' test data. For this reason, we disable
+     * because we'd then have to augment them with `buendia_patient_sync_map` records, which would
+     * couple our test integrity to the records in OpenMRS' test data. For this reason, we disable
      * {@link BaseModuleContextSensitiveTest}'s setup by putting the {@link SkipBaseSetup}
      * annotation on the class, but then we've got to explicitly init the database and authenticate
      * ourselves.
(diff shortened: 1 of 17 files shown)


openmrs/api/src/test/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAOTest.java
```diff
@@ -0,0 +1,318 @@
...
+                extractListOfUuids(results.results));
+        token = results.syncToken;
+        results = buendiaService.getPatientsModifiedAtOrAfter(token, false, 3);
+        assertArrayEquals(
+                // There should only be one in the last page.
+                Arrays.copyOfRange(EXPECTED_UUID_ORDER_NO_VOIDED_NO_DUPLICATE_TIMESTAMPS, 6, 7),
+                extractListOfUuids(results.results));
+    }
+
+
```
viniciusboson: There is one extra line here

viniciusboson:
Fantastic work, @capnfabs. 
And it's really well documented and tested.

openmrs/api/src/test/java/org/projectbuendia/openmrs/api/db/hibernate/HibernateProjectBuendiaDAOTest.java
capnfabs: Fixed.

capnfabs:
Thanks!

capnfabs:
A quick write-up of some performance aspects of this PR: https://slack-files.com/T02T5LNM4-F0FFT2L4C-c845ef2ea0

PR #42, opened 2015-11-30
Implement Paginated sync for observations in the DAO.

capnfabs:
There's two commits in this one; the second builds heavily off the first.

The first commit takes what I did for Patients and makes it generic, so exactly the same logic can be used for Observations.

The second commit just wires it all up and adds a bunch more tests :)

It's probably best to review this PR as individual commits.


llvasconcellos:
Nothing to add. Good to go!

PR #47, opened 2016-01-05
Make eventual migration to OpenMRS 1.11 easier.

capnfabs:
I was playing around with updating to OpenMRS 1.11 a few weeks ago. Whilst we can't upgrade
in our current state (we depend on an OpenMRS 1.10 DB snapshot) we _can_ fix some of the smaller
blockers.

This change fixes the test data to work in 1.11, which introduces DB constraint checks for tests,
and adds comments cross-referencing when identical versions have been defined in more than one
place.


openmrs/api/src/test/resources/org/projectbuendia/openmrs/include/obsDataSetNoDuplicateTimestamps.xml
```diff
@@ -14,20 +14,22 @@
   -->
 
 <!-- This dataset based off org/openmrs/include/standardTestDataset.xml. -->
-
+<!--
+This dataset requires samplePatientsDataSet.xml, which defines the patients referenced here.
+-->
 <dataset>
-    <obs obs_id="7" person_id="7" concept_id="5089" encounter_id="3" obs_datetime="2008-07-01 00:00:00.0" location_id="1" value_numeric="50.0" comments="" creator="1" date_created="2008-08-18 14:09:35.0" voided="false" value_coded="[NULL]" value_coded_name_id="[NULL]" value_complex="[NULL]" value_text="[NULL]" value_datetime="[NULL]" value_drug="[NULL]" uuid="fffffff"/>
```
llvasconcellos: Why no more encounter_id?

llvasconcellos:
Great! Good to go!

openmrs/api/src/test/resources/org/projectbuendia/openmrs/include/obsDataSetNoDuplicateTimestamps.xml
capnfabs: I'm not sure actually. It seems like it must have been intentional, but it's not valid to have empty `encounter_id`s, so I've restored them. I probably had removed them because there wasn't a null check constraint, but there was a foreign key check for set values, and I didn't have the encounters in the test data at the time.

PR #48, opened 2016-01-07
Allow orders to be edited and deleted.

capnfabs:
Also, adds comprehensive tests for OrdersResource.


llvasconcellos:
Good to go!

capnfabs:
Ok, please take another look at the second commit here; I switched back to an architecture closer to what Ping had implemented initially.

llvasconcellos:
OK. Only one comment but it's fine by me.

capnfabs:
Thanks!

PR #49, opened 2016-01-08
Multiple Charts

llvasconcellos:
*** THIS NEED TO BE MERGED ALONGSIDE WITH THE CLIENT PR ***
Chart divider for multiple charts.

### DIFF
--- a/tools/profile_apply
+++ b/tools/profile_apply
@@ -250,12 +250,22 @@ def apply(tabs):
         for chart_rows in split_by_title(rows):
             put_chart(chart_rows)
 
+    def clear_form(form_id):
+        """Removes all the form_fields from the given form."""
+        db.execute('update form_field set parent_form_field = null'
+                   ' where form_id = %s', form_id)  # remove foreign keys
+        db.execute('delete from form_field where form_id = %s', form_id)
+
     def apply_chart(rows, form_id):
         """Applies the chart definition given rows from the chart tab."""
+        clear_form(form_id)
         section = None
         sections = []
         section_rows = []
+
         for row in rows:
+            if row['title'].strip():
+                sections.append(('[chart_divider]', [{'label': row['title'].strip(), 'concept':'162169', 'type':''}]))
             if row['section'].strip():
                 if section and section_rows:
                     sections.append((section, section_rows))
@@ -265,12 +275,6 @@ def apply(tabs):
         if section and section_rows:
             sections.append((section, section_rows))
         apply_chart_sections(sections, form_id)
-
-    def clear_form(form_id):
-        """Removes all the form_fields from the given form."""
-        db.execute('update form_field set parent_form_field = null'
-                   ' where form_id = %s', form_id)  # remove foreign keys
-        db.execute('delete from form_field where form_id = %s', form_id)
 
     def apply_chart_sections(sections, form_id):
         """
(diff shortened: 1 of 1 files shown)


tools/profile_apply
```diff
@@ -250,12 +250,23 @@ def apply(tabs):
...
     def apply_chart(rows, form_id):
         """Applies the chart definition given rows from the chart tab."""
+        clear_form(form_id)
         section = None
         sections = []
         section_rows = []
+
         for row in rows:
+            if row['title'].strip():
+                sections.append(('[chart_divider]',
```
capnfabs: So, if I understand this correctly:
- there's still only one `chart` section in the `profile.csv`
- each entry in the `title` adds a new `chart_divider` to the chart form in OpenMRS
- When this is parsed in the client, the chart form is split up based on the `chart_divider`s?

tools/profile_apply
```diff
@@ -284,7 +289,10 @@ def apply(tabs):
         clear_form(form_id)
 
         for i, (section, rows) in enumerate(sections):
-            section_type = section[:1] == '[' and 'TILE_ROW' or 'GRID_SECTION'
+            if section == '[chart_divider]':
+                section_type = 'CHART_DIVIDER'
+            else:
+                section_type = section[:1] == '[' and 'TILE_ROW' or 'GRID_SECTION'
```
capnfabs: This syntax is super confusing to read. I'm pretty inexperienced in python, is this what people use instead of a ternary operator?

capnfabs:
I'd prefer that we did this as separate charts, but because that would require a rethink of the way that we indicate something is a chart, I'm ok with your approach for the time being.

One more thing to test: will this work if the profile author writes `chart` in the `tab` column for each chart?

Ok to merge after addressing these comment :+1:

llvasconcellos:
This is the profile I'm currently testing things.

[nutrition-chad16-fr-2charts-v2.txt](https://github.com/projectbuendia/buendia/files/88342/nutrition-chad16-fr-2charts-v2.txt)

tools/profile_apply
llvasconcellos: * No each chart should have it's "chart" section. It would work without it (and so do forms), but only if you have them one after the other.
* Yes, the title is the real divider, not the "chart" (as it is to forms).
* Yes

tools/profile_apply
llvasconcellos: Yes. It took me a while to understand it also. 
The *section[:1]* means the first char from section variable.

llvasconcellos:
@capnfabs After I studied the implementation of our chart I realized that there is no problem this implementation as a definitive one for the server side (the client side can be improved later).
On the server side, a chart is a form called "Chart display order". To have multiple forms, one for each chart, would mean that we needed to maybe create a more database structure (and much more code) top relate them.
I really think that, for the server side, this is a clean, minimalist and yet elegant solution for the problem.
For the client side is another story. There we have more flexibility, so we should re-structure the database to accommodate multiple charts. Adding to that I would refactor our chart related classes because it is starting to become messy. I didn't do any of that because of time constrictions and that's why I said to you that I'm not proud of the (client's) solution.

capnfabs:
My concern on the server side is that having multiple charts in the same OpenMRS form is another example where we don't conform to OpenMRS's data model. That's generally been something that I want to avoid, because longer-term, I think we've got potential to position this as something that other users of OpenMRS can use.

But yeah, for the time being, I'm ok with this on the server. And, I think the client side scrubbed up pretty well, actually :)

llvasconcellos:
@capnfabs Thanks!

PR #50, opened 2016-01-12
Support multiple servers on same host.

capnfabs:
- `tools/openmrs_setup`, `tools/openmrs_build` and `tools/openmrs_run` all support having an
  environment variable `OPENMRS_SERVER` defined, which specifies the server to setup, build and run.
- `OPENMRS_SERVER` defaults to `"server"`, which retains compatibility with existing behavior.
- The database name `"openmrs_$OPENMRS_SERVER"` is now used, instead of just `openmrs`.
- `tools/openmrs_run` now supports a `-p` parameter to specify a port to run on, and a `--debug`
  flag, which enables debugging. By default, `-p` is `9000` (same as before), and `--debug` is off
  (behavior has changed).

### DIFF
--- a/tools/openmrs_run
+++ b/tools/openmrs_run
@@ -12,17 +12,51 @@
 
 # Starts a local OpenMRS server for development.
 
+## Usage:
+##   tools/openmrs_run [-p PORTNUM] [--debug]
+##
+## Note that it's possible to change the OpenMRS server ID by setting OPENMRS_SERVER.
+
 set -e
-SERVER_ID=server
+
+PORT="9000"
+DEBUG=0
+
+while [[ $# > 1 ]]
+  do
+  key="$1"
+
+  case $key in
+    -p)
+      PORT="$2"
+      shift # past argument
+      ;;
+    --debug)
+      DEBUG=1
+      ;;
+  esac
+  shift # past argument or value
+done
+
+
+if [ -z "$OPENMRS_SERVER" ]; then
+  SERVER_ID="server"
+else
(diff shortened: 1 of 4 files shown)


dancunningham:
Is this to facilitate development on server-server sync?

capnfabs:
Yes.

capnfabs:
Note that this doesn't work with the profile_apply script.

tools/openmrs_run
```diff
@@ -12,17 +12,51 @@
...
+##
+## Note that it's possible to change the OpenMRS server ID by setting OPENMRS_SERVER.
+
 set -e
-SERVER_ID=server
+
+PORT="9000"
+DEBUG=0
+
+while [[ $# > 0 ]]
```
llvasconcellos: Out of curiosity: What this loop does?

llvasconcellos:
OK. No problems. Now I'll run tools/openmrs_run --debug

tools/openmrs_run
capnfabs: Loops until there are no unprocessed command-line arguments. `$#` is the number of command line arguments. Every time we call `shift` the first argument is removed from the set, so `$#` is decremented.

capnfabs:
Not going to merge this, it got pretty complex passing the database connection properties from OpenMRS into python.

tools/openmrs_run
llvasconcellos: Interesting. Thanks!

capnfabs:
Abandoning this.

PR #51, opened 2016-01-14
Truncate form UUIDs to 38 chars in `profile_apply`.

capnfabs:
Fixes https://github.com/projectbuendia/client/issues/198.

### DIFF
--- a/tools/profile_apply
+++ b/tools/profile_apply
@@ -7,6 +7,7 @@ import os
 import re
 import sys
 import uuid
+import warnings
 
 CHART_UUID =  'ea43f213-66fb-4af6-8a49-70fd6b9ce5d4'
 LOCALE = 'en_GB_client'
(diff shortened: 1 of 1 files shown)


llvasconcellos:
This could, at some point, cause problems. Can we at least output a warning message?
I know the possibility is remote but someone, someday, might have two different forms with the first exact 38 chars.

dancunningham:
I think it did this before and caused a problem in that it meant forms with similar names ended up with the same UUID.

Is that somehow dealt with?

tools/profile_apply
```diff
@@ -339,11 +339,18 @@ def apply(tabs):
     def normalize_name(name):
         return re.sub('[\W_]+', '_', name.lower()).strip('_')
 
+    def truncate_for_uuid(uuid):
+        """
+        UUIDs are only allowed to be 38 characters. Truncate the given string
+        to this length.
```
dancunningham: and then do something to guarantee uniqueness!

capnfabs:
It's a problem if two forms have the same first `38 - len("buendia-form-") = 25` chars, which is more likely than the first 38 :) But still not super likely.

We can't do a lot to guarantee uniqueness at the moment because we rely upon two specific UUIDs that are generated in this way on the client side. See `PatientChartController#OBSERVATION_FORM_UUID` and `PatientChartController#EBOLA_LAB_TEST_FORM_UUID`. I don't think those are used anymore, though, so I think we could probably remove them.

Once we do that, there's nothing special about the `"buendia-form-"` prefix - it's not used anywhere else in code.

Then, we need to ensure that the UUIDs are deterministic, so that modifying a form and re-applying the profile overwrites the existing one. I think the best way to do this would be to hash the chart name and put that in the UUID field.

Having said all of that, let's not worry about all that for the time being. I'll add a warning.

capnfabs:
Ok, I added a super user-friendly warning message as @llvasconcellos suggested.

PR #56, opened 2016-01-18
Fix CSV Export

### LINKED ISSUE #53: Fix CSV Export
Make sure it exports all encounters and observations
Make sure it includes key required data including patient ID, name, age and sex

llvasconcellos:
Fixed patient birth date bug and added try/catch for non stop exporting.
This fixes #53


llvasconcellos:
* There were a problem with the exporting when the patient had no birth date: this would throw and exception and stop exporting. I added a null verification for the date.
* I also added a try / catch for each line of the csv exporting so if any error occur only the problematic line will be skipped and the export will continue.
* Another change added was to clean the line before adding a new one. This prevents repeating a value from the preview line (like the birth date).

PR #57, opened 2016-01-18
Allow clients to send a provider UUID with an Xform to attribute the …

capnfabs:
…form to a user.

This is a partial fix for https://github.com/projectbuendia/client/issues/214.

PR #58, opened 2016-01-18
Attribute Orders to the users that created them.

capnfabs:
This is a partial fix for projectbuendia/client#214.

Note that this PR depends on work from pull request #57, only the latest commit 7936b34 should be reviewed as part of this pull request.


llvasconcellos:
Looks good!

PR #59, opened 2016-01-18
Add the enterer's UUID to observations sent from the server.

capnfabs:
This allows clients to properly attribute observations to a specific user.

Note that we're especially carefully not to crash sync if an observation was recorded by someone
that doesn't exist in a way that's compatible with the Buendia model of a 'User'.

Each Buendia user has the following three objects in OpenMRS:
- A Person, which contains information such as the person's name
- A User, which contains login credentials (though Buendia doesn't really populate these properly)
- A Provider, which is the OpenMRS representation of a person who administers orders.

When users are synced from the server to the client, the Provider ID gets sent down at the moment,
so we have to do a bit of work to convert providers into users and back throughout.

### DIFF
--- a/openmrs/omod/src/main/java/org/openmrs/projectbuendia/Utils.java
+++ b/openmrs/omod/src/main/java/org/openmrs/projectbuendia/Utils.java
@@ -12,8 +12,13 @@
 package org.openmrs.projectbuendia;
 
 import org.openmrs.Order;
+import org.openmrs.Person;
+import org.openmrs.Provider;
+import org.openmrs.User;
+import org.openmrs.api.context.Context;
 import org.openmrs.projectbuendia.webservices.rest.InvalidObjectDataException;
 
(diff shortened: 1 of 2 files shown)


openmrs/omod/src/main/java/org/openmrs/projectbuendia/Utils.java
```diff
@@ -183,4 +189,18 @@ public static Order getRootOrder(Order order) {
...
         return order;
     }
+
+    public static @Nullable Provider getProviderFromUser(@Nullable User user) {
+        if (user == null) {
+            return null;
+        }
+        Person person = user.getPerson();
+        Iterator<Provider> providers =
+                Context.getProviderService().getProvidersByPerson(person).iterator();
```
llvasconcellos: Soo easy!

llvasconcellos:
Great work! Good to go!
:metal:

PR #60, opened 2016-01-19
Fix patients not syncing until an observation was added.

capnfabs:
Patients weren't able to be synced until they were edited, due to a bug in the SQL trigger that
creates entries in the `buendia_patient_sync_map` on person insert.

This change fixes that bug. Fixes https://github.com/projectbuendia/client/issues/187


llvasconcellos:
:ok_hand: Looks good!

llvasconcellos:
Every time a new patient is created, an observation is added:
	Date of admission to Ebola Treatment Unit (ETU)
Shouldn't the person already be a patient by now?

capnfabs:
Yeah, the weird thing is that it used to work fine - I tested and implemented this by just adding a huge number of patients using `curl`, and then synced them from the tablet without any problems. Either way, this fix is more correct, so I'm still happy to make it.

capnfabs:
To be clear, it's not that there's a stage when the person isn't a patient, it's that the old trigger runs immediately after the `person` record is stored but immediately before the `patient` record is stored, which is part of the same atomic transaction.

PR #61, opened 2016-01-19
Sync secondary concepts that aren't referenced anywhere else.

capnfabs:
Previously, secondary concepts in chart fields (any concepts other than the first listed concept)
were not synced to the client. This resulted in the client not knowing how to render these secondary
concepts unless they were also a primary concept in another chart field.

Fixes https://github.com/projectbuendia/client/issues/202.

### DIFF
--- a/openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/ConceptResource.java
+++ b/openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/ConceptResource.java
@@ -25,9 +25,13 @@ import org.openmrs.module.webservices.rest.web.RequestContext;
 import org.openmrs.module.webservices.rest.web.annotation.Resource;
 import org.openmrs.module.webservices.rest.web.representation.Representation;
 import org.openmrs.projectbuendia.ClientConceptNamer;
+import org.openmrs.projectbuendia.Utils;
 import org.projectbuendia.openmrs.webservices.rest.RestController;
 
+import javax.annotation.Nullable;
+import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/ConceptResource.java
```diff
@@ -107,6 +118,48 @@ protected Concept retrieveImpl(String uuid, RequestContext context, long snapsho
...
+        }
+        List<Concept> returnValue = new ArrayList<>();
+        try {
+            SimpleObject extendedData = SimpleObject.parseJson(fieldDescription);
+            Object object = extendedData.get("concepts");
+            if (! (object instanceof List)) {
+                return Collections.emptyList();
+            }
+
+            List list = (List) extendedData.get("concepts");
```
dancunningham: Can this just be `List list = (List) object;` ?

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/ConceptResource.java
```diff
@@ -96,10 +100,17 @@ protected Concept retrieveImpl(String uuid, RequestContext context, long snapsho
...
+                    ret.add(fieldConcept);
+                    for (ConceptAnswer answer : fieldConcept.getAnswers(false)) {
+                        ret.add(answer.getAnswerConcept());
+                    }
+                }
+                // Fetch and add additional concepts that might have been stored in the field
+                // description.
+                for (Concept additionalConcept :
+                        getConceptsFromFieldDescription(field.getDescription())) {
+                    ret.add(additionalConcept);
```
dancunningham: For (most) fields that are just a single concept, will it always have added the concept already on line 106 and try to add it again to the HashSet here?

If so, could lines 102-108 be removed?

Though maybe it's more intuitive as written, and then works in case we ever have a field that doesn't include its concept id in the description?

dancunningham:
Overall looks good to me; I'll test it once it's merged into dev and deployed by jenkins.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/ConceptResource.java
capnfabs: Good catch! That's the whole reason why the variable was there in the first place :)

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/ConceptResource.java
capnfabs: For most fields, which only have a single concept, this additional code is an expensive no-op, but because we've got no way of knowing how many concepts each field references until we parse the JSON we have to live with the cost anyway. 

We can't remove the code that adds the answer concepts though, because they're not necessarily included in the JSON in the `description` field.

Does that answer your question? I'm not sure I understand.

capnfabs:
Something else to consider: also loop through any forms sent to the client, and grab their concepts too. I don't think this is necessary, though, because Xforms takes care of the display of concepts used in the forms.

PR #64, opened 2016-01-20
In profile, make yes_no really mean yes or no and introduce yes_no_unknown

dancunningham:
Changes how yes_no fields are created so they only have options yes and no, and not unknown.

Introduce new yes_no_unknown field type that allows for all three options.

### DIFF
--- a/tools/profile_apply
+++ b/tools/profile_apply
@@ -132,9 +132,10 @@ def apply(tabs):
     text_datatype = db.get('concept_datatype_id', name='Text')
     answer_datatype = db.get('concept_datatype_id', name='N/A')
     datatypes_by_type = {
-        'yes_no': coded_datatype,  # unknown (1067), no (1066), or yes (1065)
+        'yes_no': coded_datatype,  # no (1066) or yes (1065)
+        'yes_no_unknown': coded_datatype,  # unknown (1067), no (1066), or yes (1065)
         'select_one': coded_datatype,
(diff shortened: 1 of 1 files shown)


dancunningham:
I think this is pretty straightforward, but haven't tested as don't have a server set up on my machine.

@llvasconcellos can you please verify that this works by creating a profile with a yes_no field and a yes_no_unknown field, and checking they end up with the right options in OpenMRS and in the form on the tablet? (or if it looks basically OK, merge it and I can test it)

tools/profile_apply
```diff
@@ -132,9 +132,10 @@ def apply(tabs):
     text_datatype = db.get('concept_datatype_id', name='Text')
     answer_datatype = db.get('concept_datatype_id', name='N/A')
     datatypes_by_type = {
-        'yes_no': coded_datatype,  # unknown (1067), no (1066), or yes (1065)
+        'yes_no': coded_datatype,  # no (1066) or yes (1065)
+        'yes_no_unknown': coded_datatype,  # unknown (1067), no (1066), or yes (1065)
         'select_one': coded_datatype,
-        'select_multiple': coded_datatype,  # no (1066) or yes (1065)
+        'select_multiple': coded_datatype,  # set of no (1066) or yes (1065) fields
```
capnfabs: Suggest changing this comment to `# set of 'yes_no' fields `

llvasconcellos:
Perfect. Works Great!

PR #70, opened 2016-01-24
Outputs the required property of fields.

llvasconcellos:
This property was being ignored resulting all fields set to false on the client.


capnfabs:
Ok to merge :+1:

PR #71, opened 2016-01-25
Enable like search.

llvasconcellos:
This is part of the fix for the issue:
https://github.com/projectbuendia/client/issues/176


capnfabs:
Ok to merge :+1:

PR #73, opened 2016-01-25
Prevent watchdog restarting tomcat when restore is running

dancunningham:


capnfabs:
Looks good to me, merge when ready :+1:

PR #74, opened 2016-01-26
New CSV Export

llvasconcellos:
The default behavior of the new CSV Export is to merge encounters that are near each other by 30 minutes.
Each encounter within 30 minutes will be merged to the previous one overlapping the old value with the most recent one.
The interval can be customized passing a value in the url like:
http://localhost:9000/openmrs/moduleServlet/buendiadata?interval=120

### DIFF
--- a/openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
+++ b/openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
@@ -37,6 +37,7 @@ import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Calendar;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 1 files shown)


openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
```diff
@@ -89,11 +90,49 @@
     private static final int COLUMNS_PER_OBS = 3;
     private static final ClientConceptNamer NAMER = new ClientConceptNamer(Locale.ENGLISH);
 
+    public static final int DEFAULT_INTERVAL = 30;
+
+    private VisitObsValue.ObsValueVisitor stringVisitor;
+    private void defineStringVisitor() {
+        stringVisitor = new VisitObsValue.ObsValueVisitor<String>() {
```
capnfabs: Style note: generally, I prefer to put implementations like this at the bottom of the class definition, and to use
```java
private final VisitObsValue.ObsValueVisitor<String> stringVisitor = new VisitObsValue.ObsValueVisitor() {
//...
};
```

Then you don't have to have the separate method definition. You could even make `stringVisitor` static and then put the name in all caps, e.g. `STRING_VISITOR`.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
```diff
@@ -89,11 +90,49 @@
     private static final int COLUMNS_PER_OBS = 3;
     private static final ClientConceptNamer NAMER = new ClientConceptNamer(Locale.ENGLISH);
 
+    public static final int DEFAULT_INTERVAL = 30;
```
capnfabs: please always include units in the name for time-related constants, e.g. `DEFAULT_INTERVAL_MINS`.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
```diff
@@ -159,86 +214,25 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
...
+                            if ((value != null) && (!value.isEmpty())) {
                                 values[valueColumn] = value;
                                 values[valueColumn + 1] = value;
                                 values[valueColumn + 2] = value;
-                                return null;
                             }
-                        });
+                        }
+                    }
+                    if (i == (encounters.size() - 1)){
```
capnfabs: You can move the final `printer.printRecord` outside the loop and remove this conditional.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
```diff
@@ -89,11 +90,49 @@
...
     protected void doGet(HttpServletRequest request, HttpServletResponse response) throws
         ServletException, IOException {
-        CSVPrinter printer = new CSVPrinter(response.getWriter(), CSVFormat.EXCEL.withDelimiter
-            (','));
+
+        // Defines the interval in minutes that will be used to merge encounters.
+        int interval = DEFAULT_INTERVAL;
+        String intervalParameter = request.getParameter("interval");
+        if (intervalParameter != null) {
+            interval = Integer.valueOf(intervalParameter);
```
capnfabs: Maybe validate that the interval is set `> 0`?

capnfabs:
All in all, looks good.

I've left some specific comments, but also, could you add some more comments to the logic in the method? This is quite a long function now and it would be good to have a better indication of what each step is doing.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
llvasconcellos: F...k you are right. 
https://www.youtube.com/watch?v=5rAOyh7YmEc

openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
llvasconcellos: Done!

openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
llvasconcellos: Done!

capnfabs:
Nice one!

PR #52, opened 2016-01-14
Add sync module to our OpenMRS install.

capnfabs:
This allows us to sync between ITFC and ATFC servers.

DO NOT MERGE. This approach is still under evaluation.

See https://docs.google.com/document/d/18rjJkJw-GHlgPWxfah4EZtfZJ5b1DZIqYvUr4dhoWro/edit?usp=drive_web for more options we're considering.

PR #75, opened 2016-01-26
Add sync module to the dev server and buendia-server package.

capnfabs:
This supersedes #52.

Tested both on local dev server and by flashing to an Edison.

Note that we're currently building against the 1.3-SNAPSHOT version of OpenMRS' sync module, which includes a bugfix that I implemented in order to allow orders to sync (https://github.com/openmrs/openmrs-module-sync/pull/32).


.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "db-snapshot"]
 	path = db-snapshot
 	url = ../db-snapshot
+[submodule "third_party/openmrs-module-sync"]
+	path = third_party/openmrs-module-sync
+	url = git@github.com:openmrs/openmrs-module-sync.git
```
llvasconcellos: @capnfabs:
What is the advantage of creating this as a git submodule? openmrs-module-xforms is not treated the same.
Shouldn't we avoid breaking things with OpenMRS updatint the module?
Can't we copy the version we are using like we did for xforms?

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
```diff
@@ -270,6 +270,7 @@ private void populateDefaultsForAllOrders(Order order) {
         // Will be null if `orderer` is null.
         User creator = Utils.getUserFromProvider(orderer);
         order.setEncounter(createEncounter(order.getPatient(), creator, new Date()));
+        order.setCreator(creator);
```
llvasconcellos: I think this is unrelated. Maybe in another PR?

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
```diff
@@ -301,8 +299,7 @@ private static String normalizeName(String name) {
 
     protected static Patient jsonToPatient(SimpleObject json) {
         Patient patient = new Patient();
-        // TODO: do this properly from authentication
-        patient.setCreator(CREATOR);
+        patient.setCreator(Context.getAuthenticatedUser());
```
llvasconcellos: I think this is unrelated. Maybe in another PR?

tools/openmrs_build
```diff
@@ -45,9 +45,17 @@ $INSTALL -DartifactId=webservices.rest -Dversion=2.6
 
 # Build the Buendia module, passing along any flags (e.g. -DskipTests)
 # given to this script.
-cd openmrs
+pushd openmrs >/dev/null
```
llvasconcellos: pushd, nice to meet you. :)

.gitmodules
capnfabs: I went for a submodule because:
- If we copy-paste the source into our directory we lose all the history
- Submodules are pinned to a specific commit - the submodule isn't updated to point at a newer commit unless you do so intentionally. So I pinned it to the version with my fix in it.

We copied xforms for different reasons, I think. It's certainly been customized a lot (See the git log for that directory).

I have to admit though, I'm not super experienced with submodules.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
capnfabs: This is actually necessary to ensure that data is consistent before and after sync. `CREATOR` was a fake `User`, that had been given ID `1` but wasn't otherwise populated. This logic relied upon the object being written to the DB and then being read from the DB again before being used.

That doesn't work with the sync module, because the sync module intercepts the record before it reaches the DB.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/OrderResource.java
capnfabs: In the previous PR this was necessary - it was a fix for the same problem as the `patient.setCreator(CREATOR)` issue I described below. Seeing as I eliminated the `CREATOR` variable in this class a while ago it probably shouldn't be necessary here.

.gitmodules
capnfabs: Ok, I'm going to copy it into the source tree instead. We'd have to change our build scripts etc to make this work, I'm reluctant to do that right now.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/webservices/rest/PatientResource.java
llvasconcellos: OK!

llvasconcellos:
Good to go!!

PR #83, opened 2016-02-04
Printable Charts

llvasconcellos:
Still a work in progress but It can be merged.


dancunningham:
@capnfabs @llvasconcellos what would need to be done to this to consider it ready to merge?

For all outstanding pull requests, I would like to have things closed off neatly - either merge it in, or document what still needs to be done to merge it in, so it's clear as we move to a period of open-source way of working.

llvasconcellos:
@dancunningham Probably we could implement it using JSP. Is that what you were waiting for @capnfabs?

capnfabs:
Yes, this is OK to merge.

PR #84, opened 2016-02-06
Printable Charts v1

capnfabs:
Merging into the pilot branch so we can try it out before piloting in ITFC.

Thanks @llvasconcellos!


capnfabs:
Note to self: merging wouldn't normally work because this branch is based on dev, but because dev has no commits that the pilot branch doesn't, we're good to go.

llvasconcellos:
:metal:

PR #85, opened 2016-02-09
Fix links in dashboard so they work without our DNS server.

capnfabs:
Some users have custom DNS servers, which ruins the whole "navigate to `server/`" thing. So instead,
we change the internal, cross-port links to point at the IP of the server (10.0.0.50), and should
inform users to try `server/` or `10.0.0.50` if that doesn't work.


llvasconcellos:
Where is defined $NETWORKING_IP_ADDRESS?

capnfabs:
It's defined in the site package. When you include `utils.sh`, the script goes through and includes everything from the `/usr/share/buendia/site` directory sorted alphabetically. `$NETWORKING_IP_ADDRESS` is defined there.

llvasconcellos:
:+1:

PR #86, opened 2016-02-10
Exclude OpenMRS logs from the backup.

capnfabs:
OpenMRS logs change frequently, and if the file changes whilst being `tar`'d, tar quits with an
error. This was causing backups to fail about 25% of the time in our pilot deployment.


llvasconcellos:
Nice catch!

PR #87, opened 2016-02-10
Strip whitespace from all cells in the profile_apply script.

capnfabs:
It was almost certainly introduced by accident, because our users are hand-writing CSV files.
Whitespace at the end of strings is indetectable to users, but breaks ODK collect's form persistence
feature, because it is trimming concept labels at some stage and expecting the (trimmed) concept
label to match the label that we supply.


llvasconcellos:
:+1:

PR #88, opened 2016-02-19
CSV Export Enhancements

llvasconcellos:
Note that now we have a ordering parameter. The default behavior is to order columns to mach the patient chart but if the ***sortByUUID=true*** parameter is added the columns will be ordered by the UUID and that's the previous behavior.

**UPDATE**
The last commit added a parameter called ***idsAndUuids***. This parameter, when true, adds the 2 extra columns for each concepts (ID and UUID). It defaults to false.

### DIFF
--- a/openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
+++ b/openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
@@ -42,10 +42,9 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.Date;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
-import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
 
@@ -125,6 +124,7 @@ public class DataExportServlet extends HttpServlet {
 
         // Set the default merge mode
         boolean merge = true;
+        boolean sort = false;
 
         // Defines the interval in minutes that will be used to merge encounters.
         int interval = DEFAULT_INTERVAL_MINS;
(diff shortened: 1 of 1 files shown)


dancunningham:
@llvasconcellos the current CSV export being done on the ATFC database on the live server in Bokoro doesn't seem to be ordering the columns as they are ordered in the chart...

Also it seems to be include some concepts that we aren't using at all e.g. "Ebola Np gene PCR CT value".

What could be going on?

Which chart definition does it get them in? Could it be getting a non-published or old chart?

The order looks almost random; it's not ordered by UUID either.

openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
```diff
@@ -158,25 +165,15 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
...
 
-        // We may want to get the observations displayed in the chart/xform, in which case there
-        // are a few
-        // sensible orders:
-        // 1: UUID
-        // 2: Order in chart
-        // 3: Order in Xform
-
-        // Order in Xform/chart is not good as stuff changes every time we change xform
-        // So instead we will use UUID order, but use the Chart form to use the concepts to display.
```
llvasconcellos: @dancunningham The above lines describe the previous column order but with this PR now the ordering is exact like the chart.

llvasconcellos:
@dancunningham This PR fix all the problems you mentioned.

llvasconcellos:
@dancunningham @capnfabs New changes where made. Please review.

dancunningham:
Looks good to me!

openmrs/omod/src/main/java/org/openmrs/projectbuendia/servlet/DataExportServlet.java
```diff
@@ -141,6 +140,16 @@ protected void doGet(HttpServletRequest request, HttpServletResponse response) t
             }
         }
 
+        // Defaults to chart order of concepts but if true the ordering will be defined by
+        // concept UUID.
+        String sortParameter = request.getParameter("sortByUUID");
+        boolean sort = (sortParameter != null) && (sortParameter.equals("true"));
+
+        // Defaults to chart order of concepts but if true the ordering will be defined by
```
capnfabs: Think you forgot to edit this comment when you copy-pasted the line below :)

PR #89, opened 2016-02-20
Chart Column Interval Config

llvasconcellos:
This is the server side implementation for the pivotal story:
***Allow profile for a chart to specify one column per encounter (or per minute)***

* Added the profile format column to the [chart_divider] field description.

This is the profile I'm using for testing it:

[nutrition-atfc-101-v3.txt](https://github.com/projectbuendia/buendia/files/150180/nutrition-atfc-101-v3.txt)

### DIFF
--- a/tools/profile_apply
+++ b/tools/profile_apply
@@ -280,7 +280,7 @@ def apply(tabs):
                          'label':           row['title'].strip(),
                          'concept':         '162169',
                          'type':            'CHART_DIVIDER',
-                         'columns_format':  row['format'].strip()
+                         'format':  row['format'].strip()
                      }]))
             if validate_str(row['type']) and validate_ints(row['concept']):
                 section_rows.append(row)
--- a/tools/profile_apply
+++ b/tools/profile_apply
@@ -274,6 +274,12 @@ def apply(tabs):
                 if section and section_rows:
                     sections.append((section, section_rows))
                 section, section_rows = row['section'].strip(), []
+
+            """
+            If the title column of the CSV is not blank we assume that
+            this row delimiters the start of a new chart so we create a new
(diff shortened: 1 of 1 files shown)


dancunningham:
Great! So this will stop notes appearing over and over again?

What is the [chart_divider] thing used for anyway? Seems like a pretty nasty hack that you're working around here. Would there be a better way?

llvasconcellos:
@dancunningham: 
"Great! So this will stop notes appearing over and over again?" What do you mean? 
[chart_divider] is used for multiple charts. See the chart order form on OpenMRS. I did it this way to deliver it faster due to time constrictions. We could do a more polished way adding more complexity.
I don't think it's nasty though - It's pretty straightforward.

capnfabs:
@dancunningham @llvasconcellos the "Notes" field was appearing in every chart for printed charts. I fixed that in the `feature/llvasconcellos-printable-charts` branch.

dancunningham:
What's the impact this particular PR is intended to have?

llvasconcellos:
@dancunningham This is part of the story I'm working on. 
***Allow profile for a chart to specify one column per encounter (or per minute)***

tools/profile_apply
```diff
@@ -276,7 +276,12 @@ def apply(tabs):
                 section, section_rows = row['section'].strip(), []
             if row['title'].strip():
```
dancunningham: Can you add a comment in the code to explain what is going on here please?

It's not clear just from the code that when a new title is found, it creates a chart divider with a dummy concept ID and the format will contain the column type and interval details.

Is it necessary to have a dummy concept ID for a section? I'm a bit uneasy about it.

dancunningham:
Now I understand what this is for :-)

PR #98, opened 2019-06-11
Ensure fully automated builds.

schuyler:
Currently, `tools/openmrs_setup` needs to be run to install the OpenMRS WAR, and this script has a [sizeable number of manual steps](https://github.com/projectbuendia/buendia/blob/dev/openmrs/README-debian.md#set-up-the-server-database-as-a-dev-site).

This PR adds:
* a cache of what are for Buendia the "default" set of Java property files for OpenMRS
* a new `openmrs_ensure_sdk` script that becomes the single point of configuration in the code base for OpenMRS SDK download
* update to `openmrs_build` to allow it to build successfully even if `openmrs_setup` has not been run
* changes to the Debian packaging Makefiles to use `openmrs_ensure_sdk` and `openmrs_build` as needed, so that `make -C packages` can be run on a fresh checkout of the codebase and the packages all get built


zestyping:
This is perfect and makes total sense.

PR #100, opened 2019-06-12
Rectify version numbering of Debian packages.

schuyler:
This branch introduces `get_package_version.sh` and includes it in `.circleci/config.yml`, as well as making it the (overridable) default in `packages/Makefile`. 

`get_package_version.sh` tries to come up with a sensible package version scheme for the current build situation. The logic we're following is that only release packages built in CI get a formal version number; all other builds should be marked as such.

Also, we should treat CI builds as more recent than release builds for the same version tag; however, we should NOT treat manual builds as more recent than either. The logic there is that we do want to allow deployment of replicable pre-release builds, but we don't want accidental deployment of
developer builds. The most recent release version is retained either way, for the sanity of the developer.

The result:

* Tagged build in CI on master: **x.y.z**
* Other CI builds: **x.y.z+bn** where _x.y.z_ is the latest tagged release on master (regardless of the currently building branch) and _n_ is the CircleCI build number
* Roll yo' own builds: **x.y.z~g1337d00d** where _1337d00d_ is an unambiguous commit hash for HEAD of the current Git branch.

This may be a lot of ado about not much but it gives us some sensible defaults and it will enable me to sleep more soundly at night.


tools/get_package_version.sh
```diff
@@ -0,0 +1,46 @@
...
+if [ -n "$CIRCLE_BUILD_NUM" ]; then
+    echo "${TAG#v}+b${CIRCLE_BUILD_NUM}"
+    exit
+fi
+
+# Otherwise, we're building locally, so the version should be "x.y.z~g<commit>"
+# where <commit> is the latest unambiguous commit identifier. This version is
+# actually treated by Debian as _less_ recent than "x.y.z".
+COMMIT=$(git describe --abbrev)
+echo "${TAG#v}~g${COMMIT#*g}"
```
zestyping: (Minor) I have a mild preference for leaving off the `g` so that the part after `~` is the commit hash; I think it's easier to visually parse.

I'm indifferent as to whether to keep or omit the `b` in front of the build number.  It's less of a problem there because build numbers don't contain letters; commit hashes can have a bunch of letters so the `g` is more confusing there.

tools/get_package_version.sh
```diff
@@ -0,0 +1,46 @@
...
+# scheme for the current build situation. The logic we're following is that
+# only release packages built in CI get a formal version number; all other
+# builds should be marked as such.
+#
+# Also, we should treat CI builds as more recent than release builds for the
+# same version tag; however, we should NOT treat manual builds as more recent
+# than either. The logic there is that we do want to allow deployment of
+# replicable pre-release builds, but we don't want accidental deployment of 
+# developer builds. The most recent version is retained for the sanity of the
+# developer.
```
zestyping: Somewhere (maybe here, maybe on the wiki or a README.md) there should be documentation of the numbering format that's optimized for readers who see these numbers but don't care how they are generated, e.g.:

> There are three kinds of package version numbers:
>     0.8.4 — This is a public release.
>     0.8.4+b15 — This is CircleCI build number 15, and the most recent preceding release is 0.8.4.
>     0.8.4~g123456 — This is a release built at Git commit number 123456, and the most recent preceding release is 0.8.4.

>

tools/get_package_version.sh
```diff
@@ -0,0 +1,46 @@
...
+# latest build number. 
+if [ -n "$CIRCLE_BUILD_NUM" ]; then
+    echo "${TAG#v}+b${CIRCLE_BUILD_NUM}"
+    exit
+fi
+
+# Otherwise, we're building locally, so the version should be "x.y.z~g<commit>"
+# where <commit> is the latest unambiguous commit identifier. This version is
+# actually treated by Debian as _less_ recent than "x.y.z".
+COMMIT=$(git describe --abbrev)
```
zestyping: When the working tree has uncommitted changes, the number should indicate that it was dirty.  Will this do what we want?

tools/README.md
```diff
@@ -77,6 +77,11 @@ A script that was never written.
 Given a directory full of `*.sql` table schemas, produces the SQL commands to
 replace all the foreign key references to `users.user_id` with `@buendia_admin`.
 
+#### `get_package_version.sh`
```
zestyping: A naming convention (that isn't completely followed! but) I would like to use is to put extensions on libraries, but no extensions on executables.  So `foo.py` is a Python module and `utils.sh` is a Bash function library, but `save_the_world` is a script you can run (and the implementation is free to switch languages).

tools/get_package_version.sh
schuyler: I'll take care of this separately

tools/get_package_version.sh
schuyler: Done

tools/get_package_version.sh
schuyler: No, but it's an extremely good idea and I've committed a change that will do that

tools/README.md
schuyler: Fixed

tools/get_package_version.sh
schuyler: https://github.com/projectbuendia/buendia/wiki/Software-Package-Management#project-version-numbering-conventions

schuyler:
Merging as all review comments have been resolved.

PR #101, opened 2019-06-15
Rebuild apt repo in projectbuendia.github.io/builds whenever a master or dev build passes.

schuyler:
This PR adds two scripts for updating an apt repo, and the CircleCI config to use them.

* `tools/index_debs` indexes a particular suite of Debian packages within a given apt repository directory
* `tools/update_apt_archive` copies Debian packages from a given source path into a particular suite in an apt repository which is also a Github repository, deduplicates the packages, re-indexes the apt repository, then commits the changes to git and pushes the whole mess up to Github.

A successful `build` job on the `master` branch of this project in CircleCI triggers an additional `apt-archive` job.  The `apt-archive` job gets the build artifacts from the previous job, checks out the latest projectbuendia/builds repo, and runs `update_apt_archive` on the `stable` suite in the apt repo.

A successful `build` job on the `dev` branch updates the `unstable` suite in the apt repo. Builds on other git branches should be ignored.

The following line can be added to an apt sources.list to access the unstable suite:
`deb [trusted=yes] https://projectbuendia.github.io/builds/packages unstable main`

CircleCI relies on an SSH key pair with read-write access to the [projectbuendia/builds](https://github.com/projectbuendia/builds) project. I've set this up as a deploy key in that project, and stored the private key in the buendia project in CircleCI. An encrypted copy of the keys has also been added as `recovery/buendia_build_keys.tar.gpg`.


schuyler:
NOTE: this branch has a bug in adding packages to the `builds/` repo and is *NOT* ready for merge

tools/update_apt_archive
```diff
@@ -0,0 +1,74 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distrib-
+# uted under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, either express or implied.  See the License for
+# specific language governing permissions and limitations under the License.
+
+set -e
+
+if [ "$1" = "-h" -o -z "$1" -o -z "$2" -o -z "$3" ]; then
+    echo "Usage: $0 <repository-dir> <branch> <artifact-dir>"
```
zestyping: It looks like $1 is supposed to be the `packages` subdirectory of the repo (as in `builds/packages`), not the repo directory itself (`builds`)—could the help message make this a tad clearer?

.circleci/config.yml
```diff
@@ -48,8 +48,72 @@ jobs:
...
+            - buendia-builds-v1-{{ .Branch }}
+            - buendia-builds-v1-
+
+      - add_ssh_keys:
+          fingerprints:
+            - "18:e2:e2:8e:12:49:a1:ca:e6:6f:e1:30:08:02:ae:43"
+
+      - run:
+          name: Get the latest builds repository
+          ### NOTE: this job must have a key that can read AND write to projectbuendia/builda
```
zestyping: Typo: `builda` -> `builds`

tools/update_apt_archive
```diff
@@ -0,0 +1,74 @@
...
+    echo
+    echo "Updates the apt archive in projectbuendia/builds with the latest"
+    echo "CircleCI build artifacts for a given branch of the"
+    echo "projectbuendia/buendia project."
+    echo
+    echo "<branch> can be either 'master' or 'dev'."
```
zestyping: Specifying the branch as an argument gives me the impression that the script uses the `<branch>` argument to determine which branch to use.  But it looks like that's entirely determined by `<artifact-dir>`; the `<branch>` argument is only translated immediately to `stable` or `unstable`.  So, for example, someone could invoke this script with the wrong `<artifact-dir>` and its output could misleadingly say it was fetching artifacts from `master` when the artifacts were actually generated on `dev` or vice versa.

If I'm understanding the script correctly, what it does is update the specified suite of a specified apt archive with the packages from a specified directory.  It doesn't actually know anything about Git branches or CircleCI artifacts.  So it seems more logical to me to specify the suite directly as an argument (`stable` or `unstable`).  I know this means that the translation from master/dev to stable/unstable has to happen somewhere else (probably `config.yml`) but it still seems better—what do you think?

.circleci/config.yml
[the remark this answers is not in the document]
schuyler: Fixed

tools/update_apt_archive
schuyler: Updated as suggested -- the suite selection code now lives in `.circleci/config.yml`

tools/update_apt_archive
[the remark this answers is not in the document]
schuyler: Fixed

schuyler:
I've tested this configuration manually using code from a future branch, and it all works: 

https://circleci.com/gh/projectbuendia/buendia/127

schuyler:
Proceeding with merge since all comments were resolved

PR #102, opened 2019-06-15
Add apt dependencies for Tomcat 7.

schuyler:
Turns out, Tomcat 7 doesn't live in any of the Debian repos we've already documented. This PR adds additional config files to `tools/apt/` (to be installed in `/etc/apt` on a stretch system) to allow `apt-get install buendia-server` to pull in the right dependencies.


schuyler:
This is basically just a documentation PR, so I'm going to go ahead and merge.

(attn @zestyping)

PR #104, opened 2019-06-17
updated instructions for recovery

Joronimo:
updated instructions for recovery


ivangayton:
This is fine, only touches the markdown instructions file in the recovery directory so extremely unlikely to cause merge conflicts. Merging.

PR #106, opened 2019-06-17
 Change diffdeb so as to ignore file timestamps.

schuyler:
This is necessary now that we're building packages in CI. CircleCI regularly makes a fresh checkout of the code base, with the result that file timestamps are different from one CI run to the next, because git doesn't set them.

This change ensures that file timestamps are never a reason in and of themselves to retain an otherwise unchanged Debian package.

This PR also adds a verbose flag to `diffdeb` for debugging purposes. This flag shouldn't be used by automated processes, since it leaves the temporary directories intact for manual investigation.

### DIFF
--- a/tools/diffdeb
+++ b/tools/diffdeb
@@ -10,20 +10,36 @@
 # OR CONDITIONS OF ANY KIND, either express or implied.  See the License for
 # specific language governing permissions and limitations under the License.
 
+verbose=$1
+if [ "$verbose" == "-v" ]; then
+    shift
+else
+    verbose=""
+fi
 a=$1
 b=$2
 if [ "$b" = "" ]; then
-    echo "Usage: $0 <file1.deb> <file2.deb>"
+    echo "Usage: $0 [-v] <file1.deb> <file2.deb>"
     echo "Exits with status 0 if two .deb files are identical in all aspects"
     echo "except for the declared version number ("Version:") in the control"
     echo "file.  Otherwise exits with status 1."
+    echo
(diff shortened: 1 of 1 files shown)


schuyler:
Merging this as it is a straightforward change and its absence is causing havoc in our builds project.

tools/diffdeb
```diff
@@ -10,20 +10,36 @@
...
+if [ "$verbose" == "-v" ]; then
+    shift
+else
+    verbose=""
+fi
 a=$1
 b=$2
 if [ "$b" = "" ]; then
-    echo "Usage: $0 <file1.deb> <file2.deb>"
+    echo "Usage: $0 [-v] <file1.deb> <file2.deb>"
```
zestyping: Please update the help string ("identical in all aspects..." etc.)

tools/diffdeb
```diff
@@ -37,16 +53,13 @@ function unpack_and_remove_version() {
     rm deb
 
     # Unzip data.tar to remove the gzip timestamp from the header
-    gunzip data.tar.gz
+    $TAR xfz data.tar.gz
```
zestyping: Please update the comments on line 46–49 and line 55.

tools/diffdeb
schuyler: committed to dev as 54f456f1

tools/diffdeb
schuyler: committed to dev as 12a180a7 (oops)

PR #121, opened 2019-06-20
Don't fail when upgrading buendia-db-init via apt-get upgrade

schuyler:
Currently, running `apt-get upgrade` when the `buendia-db-init` package is installed will cause the upgrade to fail with the following error:

```
Unpacking buendia-db (0.10.0+b222) over (0.10.0+b219) ...
Preparing to unpack .../buendia-db-init_0.10.0+b222_all.deb ...
buendia-db-init has been previously installed!
Not installing, as there is a risk of destroying user data!
If you are sure you want to wipe the database,
rm /etc/buendia-db-init-installed and reinstall.
```

While we certainly don't want to re-initialize a database automatically, we also don't want `apt-get upgrade` to fail mysteriously.

This PR resolves this issue by emitting the above message, but then skipping the database initialization, while permitting the upgrade process to return successfully. Also, this PR modifies the initialization process to record the package version in `/etc/buendia-db-init-installed`, so that future users (and installation processes) can determine the vintage of the original initialization, regardless of which `buendia-db-init` package is currently installed.


schuyler:
Tested this change manually with success.

```
buendia@buendia-staging:~$ sudo dpkg -i buendia-db-init_0.10.0+b227_all.deb
(Reading database ... 51296 files and directories currently installed.)
Preparing to unpack buendia-db-init_0.10.0+b227_all.deb ...
Unpacking buendia-db-init (0.10.0+b227) over (0.10.0+b219) ...
Setting up buendia-db-init (0.10.0+b227) ...
buendia-db-init has been previously installed!
Not configuring, as there is a risk of destroying user data!
If you are sure you want to wipe the database,
rm /etc/buendia-db-init-installed and reconfigure.
buendia@buendia-staging:~$ dpkg -s buendia-db-init |  grep '^Version:'
Version: 0.10.0+b227
```

schuyler:
The intention is that the the _default_ action when upgrading the `buendia-db-init` package is to do nothing; if a database reset is desired, then `/etc/buendia-db-init-installed` needs to be explicitly deleted before `apt-get upgrade` or `dpkg --configure buendia-db-init`.

PR #125, opened 2019-06-26
Expose port 9000 in Vagrant for testing.

schuyler:
Only applies to the canned server VM config, i.e. `vagrant up server`. Enables testers to run a server in Vagrant and connect from an Android client on the local Wi-Fi.

This PR also includes a patch to keep the Vagrant provisioned sources.list from being blown away by the Buendia Debian package install.


schuyler:
Merging without review as CI passes and this is only a change in developer tooling, not the server itself.

PR #133, opened 2019-07-01
Ensure that buendia-pkgserver / buendia-update still work.

schuyler:
In #108 I brought my changes to `tools/index_debs` into `buendia-pkgserver`, and in 9eb9b671, modified `buendia-pkgserver-import` to add the suite argument to the call to `index_debs`. However, this change broke the former script as the package repo directory was not structured with suites in separate subdirectories.

This PR fixes that update by ensuring that `/usr/share/buendia/packages` ends up with the new packages copied into the appropriate suite subdirectory, as `index_debs` now expects.

I have also (manually) confirmed that with this change, `buendia-pkgserver-import` will take a bundle ZIP, and:

1. copy any `.deb` files from the bundle into `/usr/share/buendia/packages/` and reindex them
1. copy any `NN-configuration` files into `/usr/share/buendia/site/`
1. run `commands.sh` if found

Moreover, with `buendia-update` installed, any new `.deb` files are installed according to the configuration of that package.

### DIFF
--- a/packages/buendia-pkgserver/data/usr/bin/buendia-pkgserver-import
+++ b/packages/buendia-pkgserver/data/usr/bin/buendia-pkgserver-import
@@ -75,7 +75,8 @@ function adjust_filename() {
 # packages dropped there can be moved into the repo instantaneously.
 name=$(basename $0)
 exdir=/tmp/$name.ex.$$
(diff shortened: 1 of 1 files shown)


schuyler:
Flagging @zestyping FYI... this is a small change in keeping with the design of the original code, so I'm just going to merge it

Issue #132, opened 2019-06-29
Ensure that cron doesn't fill up the mail spool


schuyler:
Debian as installed on vagrant seems to pipe a lot of stuff from cron into emails which get deposited into the local mail spool, which seems to grow by a few kB a minute. We need to make sure that this doesn't happen in production.

PR #134, opened 2019-07-01
Ensure that cron doesn't fill up root's mail spool with cron job outputs.

schuyler:
Basically add an empty "MAILTO=" line to /etc/crontab when installing buendia-utils, if no MAILTO= line has been set.

Since we can't know if this line was added by us or not on postrm, just echo a warning to that effect.

Intended as a fix to #132. Tested manually in Vagrant.

cc @zestyping for architectural review


schuyler:
Update: this patch to /etc/crontab doesn't actually work

schuyler:
This is tested and working in Vagrant, so I'm just going to merge it.

PR #139, opened 2019-07-08
Revert "Ensure that cron doesn't fill up root's mail spool with cron job outputs."

schuyler:
Reverts projectbuendia/buendia#134. I don't know why yet, but apparently this change breaks all of our cron jobs.


schuyler:
closed in favor of #140

PR #141, opened 2019-07-08
Limit the size of mail spools to 1M (with 5 rotations) using logrotate.

schuyler:
We're not running a mail server here, so this otherwise unorthodox treatment of the UNIX mail spool has the advantage of being a consistent and reliable way to ensure that `/var/mail` never fills up the disk.


zestyping:
Looks reasonable to me.

PR #143, opened 2019-07-15
Add network installation tools for a complete Buendia server

schuyler:
This PR contains the components needed to construct a *mostly automated* network installer image for the Buendia server, suitable for burning to USB drive or CD-ROM. This PR depends on #144.

## Quick Start

1. Build the installer image as described below, or download it from (somewhere).
2. Write the installer image to bootable media, using e.g.
   [balenaEtcher](https://www.balena.io/etcher/) if you're on MacOS.
3. Boot an Intel NUC or other system from the USB media. Unless the machine is
   connected to Ethernet with a DHCP server on the network, you will probably
   be prompted for networking details. This is a network installer, after all.
3. Follow the prompts related to disk partitioning and boot loader. You can hit
   enter at any prompt to use the defaults. On the NUC, this is intended to
   result in a functioning system.
4. The installer will take ~5 minutes to run, maybe longer, depending on your
   network speed, and download 300-400 MB of additional software.
5. You will be prompted to remove the installer media. When the system reboots,
   it will take about a minute to finish configuring the system. A login prompt
   may be shown before the first configuration is complete.
6. After a couple minutes, the OpenMRS console will be available at
   `http://localhost:9000/openmrs/`. If you've installed to a NUC and have a
   `buendia` Wi-Fi network, the server will be available at
   `http://10.18.0.50:9000/openmrs/`.
7. You can log into the console as either `buendia` or `root`. The password is
   the same.

## Building the Installer

Building the installer image requires a Debian(-ish) Linux system. From the
current directory, run the following:

```
  $ build-buendia-iso
```

This produces a file called `buendia-install.iso` which is a bootable ISO
image. This image can be uploaded to a different machine, or installed directly
to USB media, for example:

```
  $ sudo bash -c `cat buendia-install.iso > /dev/<usb drive>`
```

## Details

The [`build-buendia-iso`](build-buendia-iso) constructs the image by starting
with a standard Debian 9.9 'stretch' ISO, which also contains the non-free
firmware needed by the Intel NUC wireless card.

The script adds `apt-transport-https` to the list of packages to add to the
base system, and then adds a [Debian preseed file](preseed.cfg) to seed most of
the netinst prompts. This preseed configuration adds our
projectbuendia.github.io _unstable_ repository to the list of apt sources, and
then appends the following packages to the default server installation:

* `buendia-site-test`
* `buendia-server`
* `buendia-networking`
* `buendia-dashboard`

Also, the preseed configuration triggers the creation of an empty file in
`/etc/buendia-defer-reconfigure` in the new system, which suppresses most of
the Buendia-specific configruation steps until after the system has booted for
the first time and both MySQL and Tomcat are running.

Finally, the installer sets the automated curses installer to be the default,
and then rebuilds the ISO image into a new file.

### DIFF
--- a/tools/iso/preseed.cfg
+++ b/tools/iso/preseed.cfg
@@ -1389,17 +1389,25 @@ apt-listchanges	apt-listchanges/email-address	string	root
 tzdata	tzdata/Areas	select	Etc
 d-i passwd/root-password-crypted password $6$R/Jihdz0q$f2uQ7btizMXbhLhrVoqZRP17S.aHCpybPTh6wgDB1/3ShAbolz8833IFZBvm5yYvfr6XepS/3xXdVBxiKsKkT0
 d-i passwd/user-password-crypted password $6$R/Jihdz0q$f2uQ7btizMXbhLhrVoqZRP17S.aHCpybPTh6wgDB1/3ShAbolz8833IFZBvm5yYvfr6XepS/3xXdVBxiKsKkT0
+
+# This only took hours to figure out...
 # https://debian-boot.debian.narkive.com/0rQzZ4OM/bug-636145-debian-installer-installer-does-not-proceed-at-apt-setup-udeb-on-powerpc-with-preseed-cfg
 # https://forum.openmediavault.org/index.php/Thread/1265-Problem-installing-from-USB-stick/?postID=5797#post5797
 d-i apt-setup/use/netinst string
 
 # Uncertain if these options will work on the NUC
 #grub-installer  grub-installer/only_debian	boolean	true
-#grub-installer  grub-installer/choose_bootdev    select    /dev/sda
-#grub-installer  grub-installer/make_active    boolean    true
+grub-installer  grub-installer/choose_bootdev    select    /dev/sda
+grub-installer  grub-installer/make_active    boolean    true
+partman-base    partman/confirm_nooverwrite  boolean    false
+partman-base    partman/confirm_nochanges    boolean    false
+partman-base    partman/confirm              boolean    true
 
 # https://www.debian.org/releases/stretch/example-preseed.txt
 d-i apt-setup/local0/repository string [trusted=yes] https://projectbuendia.github.io/builds/packages unstable main java
 d-i apt-setup/local0/comment string Buendia Packages
-d-i pkgsel/include string apt-transport-https
-d-i preseed/late_command string bash /cdrom/install-buendia
+d-i pkgsel/include string buendia-server buendia-site-test buendia-networking buendia-dashboard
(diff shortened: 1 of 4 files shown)


schuyler:
cc @ivangayton

tools/iso/preseed.cfg
```diff
@@ -0,0 +1,1413 @@
...
+netcfg	netcfg/dhcp_failed	note	
+# No boot loader installed
+nobootloader	nobootloader/confirmation_common	note	
+# LILO installation target:
+lilo-installer	lilo-installer/manual_bootdev	string	
+# Enable source repositories in APT?
+apt-setup-udeb	apt-setup/enable-source-repositories	boolean	true
+# Select your time zone:
+# Choices: Eastern, Central, Mountain, Pacific, Alaska, Hawaii, Arizona, East Indiana, Samoa
+tzsetup-udeb	time/zone	select	US/Eastern
```
zestyping: I'm not sure if I'm reading this file right, but could we set the timezone to UTC by default?

tools/iso/preseed.cfg
```diff
@@ -0,0 +1,1413 @@
...
+# Error while reducing volume group
+# 
+# Choices: Argentina, Bolivia, Chile, Colombia, Costa Rica, Cuba, Ecuador, El Salvador, Spain, United States, Guatemala, Honduras, Mexico, Nicaragua, Panama, Paraguay, Peru, Puerto Rico, Dominican Republic, Uruguay, Venezuela, other
+d-i	localechooser/shortlist/es	select	
+# state
+# Choices: Acre, Alagoas, Amazonas, Amapá, Bahia, Ceará, Distrito Federal, Espírito Santo, Fernando de Noronha, Goiás, Maranhão, Minas Gerais, Mato Grosso do Sul, Mato Grosso, Pará, Paraíba, Pernambuco, Piauí, Paraná, Rio de Janeiro, Rio Grande do Norte, Rondônia, Roraima, Rio Grande do Sul, Santa Catarina, Sergipe, São Paulo, Tocantins
+tzsetup-udeb	tzsetup/country/BR	select	America/Sao_Paulo
```
zestyping: I really don't know what I'm talking about here, but just wondering what the es, BR, Central America stuff is about.

tools/iso/preseed.cfg
schuyler: Fixed by c23c9beb

tools/iso/preseed.cfg
schuyler: That's because the `debconf-get-selections` program which generated this file originally includes the kitchen sink, including a bunch of comments that don't make a ton of sense. If we want to put additional effort into the installer feature, we might spend a few minutes cleaning up the unnecessary options in this config file.

PR #144, opened 2019-07-18
Order the execution of scripts in /usr/share/buendia/config.d

schuyler:
In working to enable #143, we've found that many of the `preinst`/`postinst` package scripts, as well as `/usr/share/buendia/config.d` scripts expect that a Debian system is already fully running with MySQL and Tomcat installed and working. This expectation is false when installing packages from a Debian installer.

Furthermore, these scripts have relied on Debian package dependency ordering to ensure that the `config.d` scripts are run by `buendia-reconfigure` in the correct order.

We've recently made some changes (e.g. 5d935dd, 2e833f8, 1e3603e) to add a `/etc/buendia-defer-reconfigure` file flag, which causes those configuration steps across various packages which require a running database or web server to be _deferred_ until that file is removed. This removal happens in a `systemd` [one-shot service](https://github.com/projectbuendia/buendia/blob/1e3603e6e9/packages/buendia-utils/data/usr/share/buendia/systemd/buendia-reconfigure.service) which is run at next system boot. However, the consequence of these changes is that the Debian package dependency ordering no longer implicitly controls the order of execution of `config.d` scripts.

This PR corrects that lapse by adding `run-parts`-style script numbering to the scripts in `/usr/share/buendia/config.d`. Making this change also allows us to move the heart of `buendia-db-init` from package `postinst` into a `config.d` script, where it can be treated the same way as other configuration steps.

The (newly explicit) ordering of `config.d` steps is:

* 10: networking
* 30: mysql, tomcat7
* 40: db-init
* 50: openmrs
* 70: server

This PR consists of the last remaining changes to the core Buendia packages needed for the network installer implemented in #143 to work. Also, the changes make the behavior of `buendia-reconfigure` more explicit and hopefully a little clearer.


packages/buendia-utils/data/usr/bin/buendia-reconfigure
```diff
@@ -66,10 +66,11 @@ if [ -n "$1" ]; then
         reconfigure_package $package
     done
 else
-    for config in /usr/share/buendia/config.d/*; do
-        if [ -f $config -a -x $config ]; then
-            reconfigure_package $(basename $config)
-        fi
+    # Find executable scripts in config.d and sort them
+    for config in $(find /usr/share/buendia/config.d -type f -executable | sort); do
```
zestyping: Why find + sort — doesn't regular globbing with * produce the results in sorted order?  Just a minor thing — I'm curious if my understanding of globbing is wrong and there's something to learn here.

schuyler: Oops, [you are right](https://www.gnu.org/software/bash/manual/html_node/Filename-Expansion.html):

> Bash scans each word for the characters ‘*’, ‘?’, and ‘[’. If one of these characters appears, then the word is regarded as a pattern, and replaced with an alphabetically sorted list of filenames matching the pattern.

So this `find | sort` pipeline is much ado about nothing, really. I think I should put it back the way it was for clarity.

schuyler: Fixed by a7c19920

PR #187, opened 2019-08-25
Add integration tests for buendia-restore

schuyler:
This PR creates a `buendia-site-test-integration` site package to house integration tests that alter the openmrs database in a way that might be destructive.

Two tests are included: 

1. A test that creates OpenMRS locations and users, and then adds a single patient using the API. 
2. A test that uses `buendia-backup` to back up the database to an "external" (actually loopback) storage device, deletes the contents of the `patient` table, and then recovers the original state of the database using `buendia-restore`.

~This PR is a WIP due to a race condition between the restore test, and the actual cron job that it emulates, on restoring the `restore/` directory from the loopback filesystem. I'd like to squash this race condition before merging.~

Also this PR is based on #186 and should be reviewed only after that PR is merged.


packages/buendia-site-test-integration/data/usr/share/buendia/tests/60-site-test-integration-restore
```diff
@@ -0,0 +1,50 @@
...
+        return 1
+    fi
+    return 0
+}
+
+test_60_restore_from_backup () {
+    mv loop/backup.$(date +%Y-%m-%d) loop/restore || return 1
+    ( 
+        flock --verbose -w 30 -n 9 || exit 1
+        flock -u -n 9
```
zestyping: [help] I don't really understand what's happening here with `flock` (e.g. why lock and then immediately unlock, rather than hold the lock while the cron job runs?).  But is it moot because it sounds like you are planning to change or get rid of this anyway?

schuyler:
> [bug?] Oh! And one more important thing. Don't we need logic to make sure the destructive tests only run on test sites, and refuse to run on production sites? I don't think I saw that—please make sure that works before merging (and apologies if it's there and I missed it somehow).

Yes, the logic is inherent in the fact that these tests are part of the `buendia-site-test-integration` package, and not the `buendia-backup` package. So the safeguard is having to (at some point) type `apt-get install buendia-site-test-integration`...

packages/buendia-site-test-integration/data/usr/share/buendia/tests/60-site-test-integration-restore
```diff
@@ -0,0 +1,50 @@
...
+        execute_cron_right_now backup
+    ) 9>/var/run/lock/buendia-restore
```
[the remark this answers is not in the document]
schuyler: (The `flock -u` is there because `buendia-restore` will need to lock the same file itself, when it starts up deep inside a set of nested child processes)

packages/buendia-site-test-integration/data/usr/share/buendia/tests/60-site-test-integration-restore
schuyler: I got rid of it

packages/buendia-site-test-integration/data/usr/share/buendia/tests/60-site-test-integration-restore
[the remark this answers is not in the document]
schuyler: I solved this problem by moving the crontab out of the way. This way I could get rid of the weird file locking code

schuyler:
I've added a `--destructive` switch to `buendia-run-tests` which disables any test suites that declare themselves as destructive. (In this sense, "destructive" might merely mean "disruptive of running services" rather than actively destructive, but the point is that such tests should not be run on a system in production.)

This switch is a proxy for `UTIL_RUN_DESTRUCTIVE_TESTS` which can be enabled in a site configuration.

packages/buendia-site-test-integration/data/usr/share/buendia/site/40-site
```diff
@@ -17,3 +17,6 @@ MONITORING_LIMITS='
 '
 
 UPDATE_AUTOUPDATE=1
+
+# Enable destructive testing for this site by default.
+UTIL_RUN_DESTRUCTIVE_TESTS=1
```
zestyping: The convention for site settings is that they start with the package names of the things they control, so since the testing stuff is in `buendia-utils`, the `UTIL_RUN_DESTRUCTIVE_TESTS` setting should start with `UTILS_`.  I'm guessing maybe you meant this to be the case for the other `UTIL_...` variables also?

schuyler: Yes, that was an error. Fixed

schuyler:
Merging as all comments have been addressed

schuyler:
@zestyping I resolved this by following your suggestion... instead of "destructive" tests (which I didn't much care for as terminology) we now have "safe" and "unsafe" tests. "Safe" tests are more or less guaranteed not to adversely affect a production system, and they are identified with the word "safe" somewhere in the test suite filename. The `--unsafe` flag or the `UTILS_RUN_UNSAFE_TESTS` env variable must be passed to `buendia-run-tests` in order for unsafe tests to be run.

PR #198, opened 2019-09-05
Add buendia-site-reference package

schuyler:
The reference deployment consists of a NanoStation M2 on 10.0.0.1/24, providing DHCP with the DNS server set to 10.0.0.2. The wifi SSID is `buendia`.

This package configures the NUC to take 10.0.0.2/24 on its wifi interface, and provide DNS on that address.

The Wi-Fi password should be set manually by adding a `NETWORKING_PASSWORD` parameter to `/usr/share/buendia/site/75-networking` or similar.

### DIFF
--- a/packages/buendia-site-reference/data/usr/share/buendia/site/40-site
+++ b/packages/buendia-site-reference/data/usr/share/buendia/site/40-site
@@ -1,4 +1,4 @@
-# Site configuration for integration testing on an Edison.
+# Site configuration for reference deployments
 
 SERVER_OPENMRS_USER=buendia
 SERVER_OPENMRS_PASSWORD=buendia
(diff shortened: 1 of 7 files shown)


packages/buendia-site-reference/data/usr/share/buendia/site/40-site
```diff
@@ -0,0 +1,33 @@
...
+100000 /usr/share/buendia/packages
+'
+
+# Default network interfaces on the Intel NUC
+NETWORKING_WIFI_INTERFACE=wlp2s0
+NETWORKING_ETHERNET_INTERFACE=enp3s0
+
+# Join the existing "buendia" network as provided by the Nanostation
+NETWORKING_SSID=buendia
+NETWORKING_PASSWORD=buendia
```
zestyping: The password can't be "buendia", can it?  I thought we used "plumpynut" because it had to be at least 8 characters.

packages/buendia-site-reference/data/usr/share/buendia/site/40-site
```diff
@@ -0,0 +1,33 @@
+# Site configuration for integration testing on an Edison.
```
zestyping: Update this description.

packages/buendia-site-reference/data/usr/share/buendia/site/40-site
schuyler: yeahhhh

PR #209, opened 2019-09-25
Use Java options to prevent slow Tomcat startup

schuyler:
We've experienced slow server startup times in the past, due to a [fairly well-known issue](https://ruleoftech.com/2016/avoiding-jvm-delays-caused-by-random-number-generation) with Catalina blocking while looking for randomness.

In 1562e92, we tackled this issue by replacing `/dev/random` with `/dev/urandom`; however, this approach has two disadvantages: One, the PRNG devices are replaced by Debian on boot; two, it addresses a JVM "bug" by making a system-wide change with possible unintended consequences.

This PR addresses the problem a little more directly by explicitly telling the JVM which PRNG to use when starting Catalina. The use of `/dev/./urandom` (_sic_) apparently sidesteps some cleverness internal to the JVM that we evidently _don't_ want.

### DIFF
--- a/packages/buendia-server/data/usr/share/buendia/config.d/70-server
+++ b/packages/buendia-server/data/usr/share/buendia/config.d/70-server
@@ -12,16 +12,6 @@
 
 set -e; . /usr/share/buendia/utils.sh
 
-# Ensure that /dev/random points to a fast source of entropy.  If it points
-# to the blocking device, Catalina may take 5-10 minutes to start up!
-if [ ! -L /dev/random ]; then
-    if [ -e /dev/random ]; then
-        mv -f /dev/random /dev/blocking-random
-    fi
-fi
-rm -f /dev/random
-ln -sf /dev/urandom /dev/random
-
 # Provide a randomly generated password for the Buendia user, if necessary.
 generated=/usr/share/buendia/site/20-server
 if [ ! -e $generated -o -z "$SERVER_OPENMRS_PASSWORD" ]; then
--- a/packages/buendia-tomcat7/data/usr/share/buendia/diversions/etc/default/tomcat7
+++ b/packages/buendia-tomcat7/data/usr/share/buendia/diversions/etc/default/tomcat7
@@ -1,7 +1,16 @@
 TOMCAT7_USER=tomcat7
 TOMCAT7_GROUP=tomcat7
 
+# NOTE SDE 2019-09-25: the following comment says "Edison" but I see no reason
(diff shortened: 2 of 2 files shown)


schuyler:
Note: I've tested this change by installing a dev package on Vagrant and confirming that the option appears in the JVM command-line:

```
$ ps auwx | grep java
tomcat7   1744 62.0 62.2 1762996 313828 ?      Sl   22:18   1:03 /usr/lib/jvm/java-7-openjdk-amd64/bin/java -Djava.util.logging.config.file=/var/lib/tomcat7/conf/logging.properties -Djava.util.logging.manager=org.apache.juli.ClassLoaderLogManager -Djava.awt.headless=true -Xmx256m -XX:+UseConcMarkSweepGC -XX:+CMSIncrementalMode -Djava.security.egd=file:/dev/./urandom -Djava.endorsed.dirs=/usr/share/tomcat7/endorsed -classpath /usr/share/tomcat7/bin/bootstrap.jar:/usr/share/tomcat7/bin/tomcat-juli.jar -Dcatalina.base=/var/lib/tomcat7 -Dcatalina.home=/usr/share/tomcat7 -Djava.io.tmpdir=/tmp/tomcat7-tomcat7-tmp org.apache.catalina.startup.Bootstrap start
```

In _theory_ this ought to solve the issue. We will have to keep an eye out and see if the slow startup times persist.

zestyping:
If this works, it's a much better solution than the symlink.

Do we know that this command-line option causes the change we want in the OpenMRS server startup process?  For example, could you run a startup without this option, replicate the delay, then re-run it with this option and compare the log output?

schuyler:
I was able to repro the PRNG starvation issue by rebooting, confirming that Debian had replaced `/dev/random` again, and then running `while true; do time dd if=/dev/random of=/dev/null bs=1G count=1; done` in one window, while restarting the server and watching `catalina.out`. I was able to produce `Creation of SecureRandom instance ... took [367,325] milliseconds` in the log. Watching the times on `dd`, I could confirm that reads from `/dev/random` were slowing into the dozens of seconds.

With the shell loop still eating `/dev/random` I added the `CATALINA_OPTS` line from this PR, stopped and started tomcat cold, the `Creation of SecureRandom instance` line stops appearing in `catalina.out`, and start time returned to the range of 54-55000ms, which I was able to reproduce reliably.

zestyping:
Fantastic!  Thanks for the thorough testing.

PR #210, opened 2019-09-25
Run integration tests on staging after every dev build.

schuyler:
Now that we have a separate demo server, we can run the integration tests on `staging.buendia.org` on every dev build, so that we get an alert (via the CircleCI Slack integration) if the tests ever fail.

This PR splits out the CircleCI tasks of deploying new packages to the download and staging servers for clarity, and adds a final call to the staging server via SSH to run the tests.

Also included are two small fixes to the `buendia-backup` tests for previously-untested and race conditions.


schuyler:
Merging because purely test related.

schuyler:
This PR also required 343b28a -- bit of an oversight on my part previously!

PR #214, opened 2019-09-30
Fix delete patients step in backup-and-restore integration test.

schuyler:
Adds `tools/clear_server.sql` to the `buendia-backup` package so that the latter can use this script in the `60-backup-and-restore` test to reset the database state.

Incidentally also fixes an ordering issue in `clear_server.sql`.


schuyler:
purely test related, merging ahead of review

PR #216, opened 2019-10-01
Add automatic forced database upgrades when migration fails.

schuyler:
Per @zestyping's [outline](https://buendia.slack.com/archives/C02T5LNMC/p1568690670015500), this PR adds a step to `buendia-update` which completely refreshes the database on demand, if migration to the latest version is not possible.

This feature is enabled with the `UPDATE_FORCE_UPGRADE` site setting and it **will destroy whatever is in the database** after backing up to `/var/backups/buendia`. Therefore this setting should be used _only_ on demo or testing sites where the contents of the database are not valuable.

This setting is enabled by default in the `buendia-site-cloud-demo` and `buendia-site-test-integration` sites.

This PR also includes a change to CI to use `buendia-update` to upgrade packages on the staging site.

### DIFF
--- a/packages/buendia-site-cloud-demo/data/usr/share/buendia/site/40-site
+++ b/packages/buendia-site-cloud-demo/data/usr/share/buendia/site/40-site
@@ -16,7 +16,5 @@ MONITORING_LIMITS='
 100000 /usr/share/buendia/packages
 '
 
-# Enable automatic upgrades and forced upgrades, so that we always have the
-# latest code running.
+# Enable automatic upgrades!
 UPDATE_AUTOUPDATE=1
-UPDATE_DESTRUCTIVE_MIGRATION=1
--- a/packages/buendia-site-cloud-demo/data/usr/share/buendia/site/40-site
+++ b/packages/buendia-site-cloud-demo/data/usr/share/buendia/site/40-site
@@ -16,4 +16,7 @@ MONITORING_LIMITS='
 100000 /usr/share/buendia/packages
 '
 
+# Enable automatic upgrades and forced updates, so that we always have the
+# latest code running.
 UPDATE_AUTOUPDATE=1
+UPDATE_FORCED_UPGRADE=1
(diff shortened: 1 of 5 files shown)


packages/buendia-update/data/usr/bin/buendia-update
```diff
@@ -54,11 +50,47 @@ if [ -n "$1" ]; then
...
     packages=$(echo $(sed -e 's/#.*//' /usr/share/buendia/packages.list.d/*))
 fi
+
+# Attempt database migration first, if requested
+if echo $packages | grep buendia-server; then
+    echo "--- Attempting database migration"
+    apt-get -V -y --allow-unauthenticated install buendia-db-migrate
+
+    from_version=$(dpkg-query --showformat='${Version}' --show buendia-server)
+    to_version=$(dpkg-query --showformat='${Version}' --show buendia-db-migrate)
```
zestyping: The destination version is the version of the `buendia-db` package that is about to be installed.  Is there a possibility that `to_version` will be wrong because it's the latest version of `buendia-db-migrate`, rather than the latest version of `buendia-db`?

packages/buendia-site-cloud-demo/data/usr/share/buendia/site/40-site
```diff
@@ -16,4 +16,7 @@ MONITORING_LIMITS='
 100000 /usr/share/buendia/packages
 '
 
+# Enable automatic upgrades and forced upgrades, so that we always have the
+# latest code running.
```
zestyping: On demo, we don't want demo data to be blown away without manual intervention; forced upgrade should be possible (and easy) but not automatic.

packages/buendia-update/data/usr/bin/buendia-update
schuyler: (Can we think of a circumstance where they would be different?)

schuyler: Part of the problem with the question is that there's no easy way to know what version of a package "would" Debian install; it just installs it. Having separate `db-migrate` versus `db` packages helps this a bit because it allows the `db` package to ensure that the latest helper package can do its thing.

schuyler: We could parse the output of `apt-cache policy` but this doesn't seem to be documented nor intended for machine-readable consumption.

zestyping: Yeah, this doesn't seem likely to cause a problem in practice.  I think this is fine for now.

PR #217, opened 2019-10-01
Fix Wi-Fi watchdog behavior on NUC.

schuyler:
Currently `buendia-wifi-watchdog` uses `wpa_cli status` to check the status of the wireless link; however, for unknown reasons, this command never reports that the Wi-Fi is connected on the NUC, even when it definitely is.

The consequence of this behavior was `buendia-networking` was causing the wireless connection to drop on the NUC for about 7-8 seconds out of every minute.

This PR changes `buendia-wifi-watchdog` under the hood to perform two checks each time it runs:

1. Run `iwconfig` and see if the `NETWORKING_WIFI_INTERFACE` is associated to an access point
2. If so, find the broadcast address for that interface, and ping it to see if any other hosts respond

If either of these checks fails, then `buendia-wifi-watchdog` will reconfigure the networking stack, which should cause the network to come back. I have tested both conditions manually on the NUC and both times the Wi-Fi came back online within 70 seconds as anticipated. Otherwise, when operating normally, the aforementioned network dropouts disappear as expected.

### DIFF
--- a/packages/buendia-networking/data/usr/bin/buendia-wifi-watchdog
+++ b/packages/buendia-networking/data/usr/bin/buendia-wifi-watchdog
@@ -43,26 +43,30 @@ if bool "$NETWORKING_AP"; then
     exit 1
 fi
 
-if [ -z "$NETWORKING_SSID" ]; then
-    echo "Not checking wifi: NETWORKING_SSID is unset."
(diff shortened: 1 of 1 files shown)


zestyping:
I discovered and fixed this yesterday.

schuyler:
🤷‍♂

PR #218, opened 2019-10-01
Refactor networking configuration and disable Wi-Fi power management

schuyler:
This PR refactors `config.d/10-networking` to remove some cruft and improve flexibility.

In particular, don't just assume that we always have a Wi-Fi interface. If we *do*, then treat it as the main interface for talking to clients, and treat the ethernet as a possible backhaul connection. If we *don't*, then treat the ethernet connection as the main interface for talking to clients.

This changes makes it easier to disable Wi-Fi power management, which has been causing local network ping times to the wireless interface on the NUC to go up to 100-500ms. With this change, ping times to the NUC tend to be under 10ms, and rarely go higher than 50ms.


schuyler:
Note: the network ping times are still _somewhat_ flaky with this change... I would expect that ping times on a local network would _never_ get above 10ms, ever. So there may still be some tuning that we want to do even after this.

Issue #221, opened 2019-10-09
buendia-reconfigure networking (with no changes) can bring down the wifi connection


zestyping:
On a reference deployment, I ran `buendia-reconfigure -f networking` and it brought down networking even though the configuration hadn't changed.

As I remember fuzzily from the time, it seemed that `NETWORKING_DHCP_DNS_SERVER` was 1, which caused the script to take the "Turning on DHCP and DNS server; disabling DHCP client" path.  But we still want to be a DHCP client, because the WiFi AP is the DHCP server in our reference setup.

I think in this state, everything (including DHCP) works properly on boot and the server gets an address, but attempting to reconfigure while up relinquishes the IP address.

schuyler:
I understand this to be the intended behavior for `buendia-reconfigure -f networking` -- the `-f` flag forces reconfiguration. The main interface needs then to be brought down so that it can definitely be brought up with the right settings.

The script shouldn't disable dhcpcd unless we have a `NETWORKING_DHCP_RANGE` set, as described in the [networking documentation](https://github.com/projectbuendia/buendia/blob/dev/packages/buendia-networking/README.md). I'll submit a PR that fixes this.

zestyping:
It is not the intended behaviour for `buendia-reconfigure -f networking` to (finally) result in a change to the networking configuration of the machine if the network settings in the `site` directory have not changed.  It may interrupt the network briefly during the reconfiguration process, but it should not bring down the network permanently (which is the incident that this issue is about).

schuyler:
Ok, my forthcoming PR should fix that, at least.

PR #228, opened 2019-10-12
Improvements to networking logic

### LINKED ISSUE #219: DHCP request causes a 60-second delay on startup when no Ethernet cable is connected
In the typical deployment setup, the server connects to the network via WiFi and not Ethernet.  All server startup steps wait until the networking setup is complete, so for 60 seconds everything is blocked.

schuyler:
This PR provides a number of improvements to the existing `buendia-networking` config.d logic.

1. Use of `allow-hotplug` on the Ethernet interface defers DHCP client requests until an Ethernet cable is plugged into the adapter. This prevents a ~1m delay to system startup when no Ethernet cable is connected.
2. Rather than explicitly enabling or disabling the DHCP client daemon in the config.d script, we just let `ifup` deal with running `dhcpcd` or not.
3. Add a `NETWORKING_IP_FALLBACK` setting that provides a fallback IP address for the Ethernet adapter in the event that a cable is plugged in but no DHCP server is found. This ensures that an admin can connect a laptop and SSH to port 9022 on a known static IP address, no matter what.
4. This PR reverts the adoption of separate files in `/etc/network/interfaces.d/*` provides in PR #218. The justification is that, if a device is removed e.g. by setting `NETWORKING_ETHERNET_INTERFACE` to "", the `interfaces.d` approach would leave Buendia with no way of knowing which interfaces were removed. Having a single interfaces file means we simply overwrite any configuration changes and have done.

This PR fixes #219 and #221.


zestyping:
This seems like exactly what we want.  Great!

PR #234, opened 2019-10-31
Revert PR #230 as authenticating under CrawlerSessionManagerValve opens a security risk

schuyler:
The IP-address-based Tomcat session reuse enabled by
`CrawlerSessionManagerValve` in #230 unfortunately permits the following scenario:

1. Client A uses HTTP Basic Auth to authenticate to the API.
2. Client A departs from the network.
3. Client B connects to the API within 5 minutes, using the same IP address and user agent string as Client A, but without authenticating.
4. Client B is assigned Client A's session by `CrawlerSessionManagerValve`
and is now using Client A's credentials.

I think we can agree that this is not what we intended. Unfortunately I think it means we need to reopen #225 and come up with a more robust solution.


zestyping:
Alas.

Issue #226, opened 2019-10-09
Server clock gets reset to 2016-01-01 sometimes


zestyping:
The result is catastrophic, causing clients to sync forever, running up load on the server, causing it to run out of memory, and preventing clients from ever reaching a stable state.

The message log shows that during boot, the CMOS clock is read as 2016-01-01, and then the clock is adjusted to 2016-11-03 at a later point during startup.

zestyping:
This does not happen on every boot; it appears to be random.

zestyping:
Here's the output of `grep 'setting system clock' messages`.  The lines are sequential even though the timestamps are not (because they come from the sometimes incorrect system clock).  You can see that the clock sometimes tracks the correct date and time and sometimes gets reset to 2016-01-01 or 2016-11-03.

```
Nov  4 17:05:16 buendia kernel: [    1.061219] rtc_cmos 00:03: setting system clock to 2016-11-04 17:05:13 UTC (1478279113)
Oct  2 16:21:20 buendia kernel: [    1.051241] rtc_cmos 00:03: setting system clock to 2019-10-02 16:21:17 UTC (1570033277)
Nov  3 17:16:42 buendia kernel: [    1.057654] rtc_cmos 00:03: setting system clock to 2016-01-01 00:00:15 UTC (1451606415)
Nov  3 17:37:54 buendia kernel: [    1.049988] rtc_cmos 00:03: setting system clock to 2016-11-03 17:37:51 UTC (1478194671)
Oct  3 17:00:21 buendia kernel: [    1.048384] rtc_cmos 00:03: setting system clock to 2019-10-03 17:00:18 UTC (1570122018)
Oct  3 20:13:26 buendia kernel: [    1.050928] rtc_cmos 00:03: setting system clock to 2019-10-03 20:13:24 UTC (1570133604)
Oct  3 22:43:55 buendia kernel: [    1.058551] rtc_cmos 00:03: setting system clock to 2019-10-03 22:43:52 UTC (1570142632)
Oct  4 07:47:08 buendia kernel: [    1.058335] rtc_cmos 00:03: setting system clock to 2019-10-04 07:47:05 UTC (1570175225)
Nov  3 17:16:42 buendia kernel: [    1.049575] rtc_cmos 00:03: setting system clock to 2016-01-01 00:00:15 UTC (1451606415)
[...]

PR #238, opened 2019-11-19
Check the hwclock on boot and disable the Buendia API if it has moved backwards since the last shutdown.

zestyping:
Closes: #226 

If the internal battery power is lost, it will reset the hardware clock.  To prevent the resulting catastrophic failure (which we experienced once during a demo in Bunia), this PR checks for this condition on boot and disables the Buendia API.  This doesn't automatically fix the problem, but at least we can fail safely and obviously so that a human will be prompted to investigate immediately.

### DIFF
--- a/packages/buendia-utils/data/usr/bin/buendia-init
+++ b/packages/buendia-utils/data/usr/bin/buendia-init
@@ -1,0 +1,63 @@
+#!/bin/bash
+
+LOG=/var/log/buendia/buendia-init.log
+HWCLOCK_FILE=/var/log/buendia/hwclock.final
+
+# Ensure no dependency on any environment variables, even PATH.
+export PATH=/sbin:/usr/sbin:/bin:/usr/bin
+
+action="$1"
+
+# Reports the date and time on the system clock to nanosecond precision.
+function nanodate() {
+    date +'%Y-%m-%d %H:%M:%S.%N'
+}
+
+# Saves the time reported by the hardware clock (for use during shutdown).
+function save_hwclock() {
+    hwclock -u > $HWCLOCK_FILE
+}
+
+# Checks the hardware clock to see if it has moved backwards since the last
+# shutdown, and if so, disables the Buendia API to prevent database corruption.
+#
+# Loss of battery power to the internal hardware clock can cause the clock to
+# reset to a time in the past.  If data is written to the database that is
+# timestamped in the past, it is not only invalid, but can also make Buendia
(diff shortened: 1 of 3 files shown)


packages/buendia-utils/data/usr/bin/buendia-init
```diff
@@ -0,0 +1,63 @@
...
+# reset to a time in the past.  If data is written to the database that is
+# timestamped in the past, it is not only invalid, but can also make Buendia
+# clients unusable.  Hence, in this eventuality, we disable the Buendia API.
+function check_hwclock() {
+    now=$(echo $(/sbin/hwclock -u))
+    final=$(echo $(cat $HWCLOCK_FILE))
+
+    if [[ "$now" < "$final" ]]; then
+        systemctl mask tomcat7
+        buendia-led pink 100  # system has entered an error state
```
schuyler: @zestyping does `buendia-led` do anything anymore?

packages/buendia-utils/data/usr/bin/buendia-init
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+LOG=/var/log/buendia/buendia-init.log
+HWCLOCK_FILE=/var/log/buendia/hwclock.final
```
schuyler: What does the `.final` suffix signify here?

schuyler:
I have a couple specific questions in separate comments.

However the overall intention of this PR is clear, the architectural choices seem sound, and I can't find any obvious bugs. I think it's a solid resolution of the issue.

If you've tested this locally, I think it's reasonable to go ahead and merge.

packages/buendia-utils/data/usr/bin/buendia-init
zestyping: Not at the moment.  These `buendia-led` calls crept in because I have a local experiment that controls the NUC LED and it's pretty sweet, but isn't ready for production yet.

`buendia-led` is written to exit with success no matter what, so it's harmless here.

packages/buendia-utils/data/usr/bin/buendia-init
zestyping: That this file was written during shutdown.  (In systemd this target is called `final`.)  Maybe it's cryptic, though; do you think it should be named `hwclock.shutdown` or `hwclock.last` or something better?  Happy to rename.

zestyping:
Thanks for the quick review!

PR #239, opened 2019-11-19
Fix errors in buendia-backup when certain config paths don't exist yet.

### LINKED ISSUE #207: buendia-backup expects /usr/share/buendia/distilled to exist
```
++ buendia-log buendia-backup /var/backups/buendia 200000
2019-09-17 19:08:30 19413> (start) buendia-backup /var/backups/buendia 200000
2019-09-17 19:08:30 19413> Backing up to /var/backups/buendia/backup.2019-09-17...
2019-09-17 19:08:31 19413>
2019-09-17 19:08:31 19413> Saving Buendia configuration...
2019-09-17 19:08:31 19413> tar: Removing leading `/' from member names
2019-09-17 19:08:31 19413> tar: /usr/share/buendia/distilled: Cannot stat: No such file or directory
2019-09-17 19:08:31 19413> tar: Exiting with failure status due to previous errors
2019-09-17 19:08:31 19413> (end, status 2)
++ mv /tmp/buendia_run_test_suite.mxb6/buendia-backup.cron /etc/cron.d/buendia-backup
```

this seems like a potential race condition with cron on a newly installed system

schuyler:
`buendia-backup` assumes that certain directories exist when the Buendia configuration is backed up, and fails with an error if not. However, if those directories don't exist, there's nothing to back up per se -- this is definitely not a fatal error and shouldn't be treated as such.

This small change to `buendia-backup` addresses this situation and fixes #207.

### DIFF
--- a/packages/buendia-backup/data/usr/bin/buendia-backup
+++ b/packages/buendia-backup/data/usr/bin/buendia-backup
@@ -122,13 +122,10 @@ sleep 1  # ensure that keep has the uniquely lowest timestamp in the directory
 
 echo
 echo "Saving Buendia configuration..."
-tar cfz "$new_dir/buendia.tar.gz" \
-  --exclude '*.omod' \
-  /usr/share/buendia/counts \
-  /usr/share/buendia/distilled \
-  /usr/share/buendia/openmrs \
-  /usr/share/buendia/profiles \
-  /usr/share/buendia/site
+# Back up only those configuration directories which already exist; don't worry
+# about any which might be missing.
+config_paths=$(ls -d /usr/share/buendia/{counts,distilled,openmrs,profiles,site} || true)
+tar cfz "$new_dir/buendia.tar.gz" --exclude '*.omod' $config_paths
 ls -l "$new_dir/buendia.tar.gz"
 
 # ---- Back up the MySQL database.


packages/buendia-backup/data/usr/bin/buendia-backup
```diff
@@ -122,13 +122,10 @@ sleep 1  # ensure that keep has the uniquely lowest timestamp in the directory
...
-  --exclude '*.omod' \
-  /usr/share/buendia/counts \
-  /usr/share/buendia/distilled \
-  /usr/share/buendia/openmrs \
-  /usr/share/buendia/profiles \
-  /usr/share/buendia/site
+# Back up only those configuration directories which already exist; don't worry
+# about any which might be missing.
+config_paths=$(ls -d /usr/share/buendia/{counts,distilled,openmrs,profiles,site} || true)
+tar cfz "$new_dir/buendia.tar.gz" --exclude '*.omod' $config_paths
```
zestyping: I think the `--ignore-failed-read` option will solve this in a simpler way.  If it's easy & quick for you to test, let's go with that, otherwise just merge this as is.  Thanks!

schuyler: It was my understanding that `--ignore-failed-read` still returns a non-zero exit value; otherwise I'd have used it

schuyler: Also, I didn't `mkdir -p /usr/share/buendia/distilled` because technically that directory belongs to `buendia-dashboard` and I thought that package should manage it.

zestyping: Okay!

PR #242, opened 2019-11-25
(Actually) exit early if we have no interfaces to configure.

schuyler:
Fixes a logic bug introduced in refactoring in 5cbf09793.


schuyler:
Merging because pure bug fix. FYI @zestyping

PR #249, opened 2019-12-03
Robustness improvements to buendia-wifi-watchdog

schuyler:
Some improvements to `buendia-wifi-watchdog` made while testing #247 in the hopes of making it more robust:

* skip running if no wifi interface is configured
* use `iwconfig` rather than `wpa_cli` to ensure that what we're looking at is network association, and ensure that `wireless-tools` is required by the package
* use `ip` to check that we actually have an IP address

Tested on the NUC by watching `buendia-wifi-watchdog.log` for several minutes, both with the network unconfigured and subsequently configured, to ensure the desired behavior. Also tested with and without a wifi interface defined. Performed same manual tests across multiple boots for confirmation.

### DIFF
--- a/packages/buendia-networking/control/control.template
+++ b/packages/buendia-networking/control/control.template
@@ -2,6 +2,6 @@ Package: ${PACKAGE_NAME}
 Version: ${PACKAGE_VERSION}
 Architecture: all
 Pre-Depends: buendia-utils
-Depends: cron-daemon, dnsmasq, hostapd, wpasupplicant, diffutils, buendia-monitoring
+Depends: cron-daemon, dnsmasq, hostapd, wpasupplicant, wireless-tools, diffutils, buendia-monitoring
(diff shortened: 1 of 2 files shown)


zestyping:
@schuyler Quick question here—why is `iwconfig` better than `wpa_cli`?

schuyler:
> @schuyler Quick question here—why is `iwconfig` better than `wpa_cli`?

The only good answer I have is that it works when `wpa_supplicant` isn't running, e.g. when the Wi-Fi network is open and doesn't require WPA security. This is not a use case the project currently has, at least not that I know of, but it was impossible for me to test the networking stack otherwise last week when the only Wi-Fi network I could use was an open one.

A less good answer is that `iwconfig` has been around longer and is more mature? But I wouldn't be sold on that myself per se.

Happy to take out this particular commit out of the PR if we're concerned about this choice.

schuyler:
@zestyping bump for review?

